PR #5, opened 2015-08-07
Add naive HLL implementations

cheddar:
This PR introduces some naive sparse, dense and compact HLL implementations.

The implementations all flow into a general framework that should enable relatively simple addition and extension of new forms of compressing the base data.

This also includes a HIP estimation mode that can work as long as the HLLs are not unioned.

@leerho @reza-iranmanesh


src/main/java/com/yahoo/sketches/Util.java
```diff
@@ -199,5 +199,16 @@ public static final String zeroPad(String s, int fieldLength) {
...
   }
-  
+
+  /**
+   * Gives the log2 of an integer that is know to be a power of 2
+   * zero.
+   *
+   * @param x number
+   */
+  public static int simpleIntlog2(int x) {
```
leerho: we don't need this method.

PR #9, opened 2015-08-12
Add some convenience methods for working with SetOperation

cheddar:
@leerho I've felt some weirdness with the casting after building a set operation and Mubeen did a couple of times too.  How do you feel about these convenience methods to help in those cases?

Btw, in adjusting the unit tests for this, it became clear that there aren't any unit tests for `ANotB` with a nominal entries different from the default.  Also, there doesn't appear to be any usage of `Intersection` with the default number of nominal entries.  Neither of which are particularly alarming, but potentially worth note.

### DIFF
--- a/src/main/java/com/yahoo/sketches/hll/Fields.java
+++ b/src/main/java/com/yahoo/sketches/hll/Fields.java
@@ -11,9 +11,9 @@ package com.yahoo.sketches.hll;
  */
 public interface Fields
 {
-  public static byte NAIVE_DENSE_VERSION = 0x0;
-  public static byte HASH_SPARSE_VERSION = 0x1;
-  public static byte SORTED_SPARSE_VERSION = 0x2;
+  byte NAIVE_DENSE_VERSION = 0x0;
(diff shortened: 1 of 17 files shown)


leerho:
Thank you for the convenience methods. Good suggestion.  Now having done that you didn't always utilize the cast (comment in the code).  

I'm not too concerned about whether the tests always use the default nomEntries or not.  What this did make me think of is that I don't think I have any tests for set operations where the incoming k's are different from each other or from the target.  This is allowed in most cases, but there are some strict rules that must be followed.  I will have to look into this.  The error distribution will track that of the smallest k.

There is some other stuff in the test methods that I could look into.  For example, now that Sketches exists, there are a number of places where I could be using these methods, but don't.  I am mixed about whether to change these or not.

I haven't merged this yet.  If you could fix the unnecessary casts, then I will merge.

cheddar:
@leerho Ok, removed.

My IDE found some other simple things that I also cleaned up a bit.  Mostly just imports though.

PR #12, opened 2015-09-15
Suggested HLL Changes, primarily to HllSketchBuilder, some to Preamble.

leerho:
Includes the changes you just proposed.


cheddar:
@leerho I think you might not've pushed up as much as you think?  It appears that this is only the change to remove the comment from `HllSketch`

PR #19, opened 2015-11-19
use Math.scalb() instead of static lookup table for powers of 2

oertl:


leerho:
Thank you!  I'm going to have the author of this section take a look at this.

cheddar:
@certi Thanks for the contribution!  `Math.scalb()` is definitely safer than the `Math.pow()` we are using, thanks for this I didn't realize that the method even existed.

That said, the lookup table is there for performance reasons, looking at the `Math.scalb()` code available on the web, doing that work every time is going to add some overhead to the methods.  Given that the array is only 256 entries, and it's statically maintained (so one per classloader), I think it's better to trade-off the memory cost for the performance gain.

I'm going to close this and instead push a change that will use `Math.scalb` instead of `Math.pow` to populate the lookup table.

oertl:
Are you sure that the lookup table is faster? Math.scalb is a low cost bit manipulation operation. It just adds the second argument to the exponent of the first argument's floating point representation, wheras the lookup table also requires an implicit range check.

leerho:
Otmar,
Thanks again for your idea, you clearly had to be deep in the code to find this!  But we were skeptical that even a built-in native function call would be faster than a small, static primitive lookup table that probably resides in L2 cache.  So I was curious, so I did an A/B test of TableLookup vs Scalb():

hll-hip_tableluvsscalb

As you can see from the plot, the Math.scalb() call is about 24ns and the lookup is about 18ns in the midrange, which about 33% slower.  They both converge at the high end due to the fact that the sketch is rejecting almost everything and neither the table or scalb end up being called.  All the high numbers in the low end are due to Java warmup issues.  

This plot was run using the SketchPerformance class in the test hierarchy with these parameters:
HllSketchBuilder configuration:
LgK:	14
K:	16384
DenseMode:	false
HIP Estimator:	true
Trials Profile: LgMinTrials: 4, LgMaxTrials: 13, lgMaxU: 23, PPO: 16

You can run it yourself :)

PR #20, opened 2015-11-20
Switch to using scalb instead of pow

cheddar:


cheddar:
@leerho You want to merge this?

PR #21, opened 2015-11-21
replaced static lookup table for powers of 2 by fast on-demand calculation

oertl:
This should be now as fast as the lookup table. At least I could not measure any significant difference. I naively thought that Math.scalb() (see https://github.com/DataSketches/sketches-core/pull/19) does more or less the same, but it is more than three times slower.


leerho:
Ah, that is much better!  I did the A/B test again at it looks like this:

hll-hip_table_pow2

I ran a number of different scenarios and for all of them the difference was not noticeable. The plot above was the only one that showed about 0.5 ns difference.  But I'd declare it well within measurement error.

This is good, as it does eliminate about 2K of static storage and probably a little faster class load time.  However, I want to add some javadoc and add an assert to check the argument, as it needs to be 0 <= e <1024.

leerho:
I will merge with some other updates in a separate PR.

PR #26, opened 2015-12-16
added missing fast rejection

AlexanderSaydakov:


src/main/java/com/yahoo/sketches/hll/OnHeapCompressedFields.java
```diff
@@ -59,6 +59,7 @@ public void bucketUpdated(int bucket, byte oldVal, byte newVal) {
         exceptions_.boostrap(fields);
       }
     } else {
+      if (val <= currMin) return this;
```
leerho: This check belongs at the top of the method, not here.

Issue #33, opened 2016-02-29
Set Operations should support implied hash seeds


will-lauer:
Currently, all the set operations enforce that incoming sketches have a specified hash seed. If no hash seed is specified, the default hash seed is implied. As an option, these operations should be able to infer the hash seed from the first sketch seen, allowing the operations to be used with non-standard seeds without having to specify the seed (a secret) up front. This has the effect of moving a possible error (mismatched seeds) later in the processing and possibly hiding the error all together if only a single sketch is seen, but has the advantage that users of the set operations don't have to know what their seeds are.

leerho:
What you are requesting cannot be done safely.

The currently implemented seed "policy" is IF you decide to use a seed **different** from the default, you MUST specify the seed everywhere, for new sketches and for all set operations.  If you wish to share your unique-seed-created sketches with other organizations, AND if they need to perform set-ops, or build their own sketches to merge with your supplied data, then you must provide them with the seed through a separate secure channel.  For security reasons, your unique-seed should not be visible in the source code nor be stored along with the sketch data.   

The current sketches use a binary image Serialization-Version 3.  SerVer 3 sketches keep a 16-bit seed-hash of the chosen seed to provide a safety-check mechanism to catch the situation where a user accidentally tries to perform set operations from sketches generated with different seeds, which would otherwise silently generate garbage.  If you have old SerVer 1 sketch data (prior to Open Source and prior to the decision to allow specification of non-default seeds), then those sketch images (if generated with the correct seed), must be re-imaged using the current SerVer 3. 
This allows the safety-check mechanism to work.  This re-imaging of the old sketch data does not require access to the original data. It is just updating the sketch image to include the correct seed-hash.  The size will be the same, or in some cases, slightly smaller.

Two decisions were made to improve performance as well as improve robustness:

The Union API supports datum updates as well as sketch updates. This provides significant performance improvements especially in a real-time environments by eliminating the need for separate Update Sketches, where they would be immediately merged anyway.  The union operation cannot perform datum updates without full knowledge of what seed to use. It is NOT a good idea to just tell users they can't use the datum updates if it requires a non-default seed and they don't have the seed in hand, even if it is documented carefully.  This is because they might also have old SerVer1 sketch data that needs to be re-imaged as well.  Not forcing the users to provide the non-default seed would result in silent failures that would be difficult to debug. 

[...]

PR #39, opened 2016-03-03
Seed handling

leerho:
#33 SeedHandling.  

This PR does some code refactoring and I took a close look at the request #33.  

I did change the Sketch.wrap(Memory, seed) method so that for SerVer 1 and 2 sketches, instead of throwing an exception, it defaults to the same as if Sketch.heapify(Memory, seed) was called.  I.e., the sketch is moved to the heap.  Also, I decided to change the null behavior of Intersection in that null does not change the internal state of the intersection.  Doing this for AnotB is more complex so I'm not changing its behavior at this time.


AlexanderSaydakov:
Why null input is no longer the same as empty sketch in the intersection?

PR #41, opened 2016-03-22
Added a few tests + Clean up

jaredsburrows:
@leerho 

- Added a few tests
- Removed duplicated class `HashOperations`
- Formatting was updated to remove trailing/leading whitespaces

### DIFF
--- a/src/main/java/com/yahoo/sketches/memory/AllocMemory.java
+++ b/src/main/java/com/yahoo/sketches/memory/AllocMemory.java
@@ -4,6 +4,8 @@
  */
 package com.yahoo.sketches.memory;
 
+import java.lang.StackTraceElement;
+
 import static com.yahoo.sketches.memory.UnsafeUtil.unsafe;
 
 /**
@@ -91,11 +93,6 @@ public class AllocMemory extends NativeMemory {
     this.clear(copyToBytes, capacityBytes-copyToBytes);
   }
   
-  @Override
-  public void freeMemory() {
-    super.freeMemory();
-  }
-  
   /**
    * If the JVM calls this method and a "freeMemory() has not been called" a <i>System.err</i>
    * message will be logged.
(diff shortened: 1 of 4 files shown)


leerho:
Jared, 
Thank you for this submission!  What you have is 3 separate PRs:

1. Cleanup of trailing blanks.  This is good, and there are likely lots of other formatting improvements that could be made to bring the code up to the Google-style conventions.  Please place formatting changes in a separate PR, distinct from the more substantive changes to the code.

2. Making the JDK7 and JDK8 compatibility classes non-private to enable testing.  This is one strategy, but the other would be to use reflection to test these special classes, which is what I would prefer. The reason these were made private is because the JDK7/8 compatibility detection is really a hack, and at some point will be removed.  I don't want anyone else writing code against the library to create dependencies on this hack. If you could do this I would accept it. Please submit in a separate PR.

3. Removing the HashOperations class from tuple, add the missing methods and consolidate all HashOperations into the one in the sketches package.  This may seem logical, but the reverse of what I think needs to be done.  Only the tuple and theta packages use this class, but because of limitations in Java visibility, in order for both packages to leverage this class it must be made public.  But the methods in this class are very specialized and to use them properly requires deep understanding of the sketch algorithms.  If this class remains public, then it becomes part of the visible API of the library and others can create dependencies on these methods, which will have to be supported in perpetuity, and limits the ability of the core developers to change or remove these special methods.   My preference is to have both the tuple and theta packages have their own version of this class in their own packages, both of which could be made private.  This would have some code duplication, but for only the few specialized methods.  The other option would be to use Java security to make this class restricted.  I have not completed my investigation into this, so I am not ready to accept this portion of the PR.

jaredsburrows:
I'll submit a new PR for 1. and 2.(test to remove modifier in test and not the API itself) and leave 3. alone.

jaredsburrows:
@leerho 
1. Removed spaces
2. Fixed unit tests
3. Reverted `HashOperations`

jaredsburrows:
@leerho Since this was not merged. There are conflicts now. 

Was there any other changes you wanted to see before I fix this again?

leerho:
Don't bother just yet.  I have implemented the unit tests for the JDK7Compatibility using reflection, and I want to look at the rest of what you suggested.  

Lee.
 

    On Monday, April 25, 2016 9:28 AM, Jared Burrows <notifications@github.com> wrote:
 

 @leerho Since this was not merged. There are conflicts now. Was there any other changes you wanted to see before I fix this again?—

leerho:
The tests of the JDK7Compatibility class are no longer needed, because I have decided to drop the JDK7 compatibility mode altogether.  The cleanup of trailing blanks will be cleared up eventually when I decide on a CheckStyle configuration.  The HashOperations we already discussed.  Thank you very much for your contribution, even though it has and will be solved via other means!

PR #46, opened 2016-04-18
generic FrequentItemsSketch

AlexanderSaydakov:


src/main/java/com/yahoo/sketches/frequencies/FrequentItemsSketch.java
```diff
@@ -0,0 +1,454 @@
+/*
+ * Copyright 2016, Yahoo! Inc. Licensed under the terms of the Apache License 2.0. See LICENSE file
+ * at the project root for terms.
+ */
+
+package com.yahoo.sketches.frequencies;
+
+import static com.yahoo.sketches.Util.*;
```
b-slim: thought it is best practice is to avoid `import .*`

PR #59, opened 2016-05-05
Wide chars

AlexanderSaydakov:


src/main/java/com/yahoo/sketches/frequencies/ArrayOfStringsSerDe.java
```diff
@@ -13,15 +13,19 @@
   @Override
   public byte[] serializeToByteArray(String[] items) {
     int length = 0;
-    for (int i = 0; i < items.length; i++) length += items[i].length() + 4;
+    byte[][] itemsBytes = new byte[items.length][];
+    for (int i = 0; i < items.length; i++) {
+      itemsBytes[i] = items[i].getBytes();
```
leerho: Not specifying UTF-8 here, makes the serialization vulnerable to different encodings.

PR #60, opened 2016-05-09
fix quantiles union getResultAndReset

himanshug:
can't call getResult() as that does a copy of underlying sketch which is to be avoided.


leerho:
Right!  That is what I intended. Thanks.  Also the unnecessary imports were because I had "unused" suppressed, which I also removed.

PR #70, opened 2016-06-24
General quality fix - squid:S00122, squid:UselessParenthesesCheck, squid:S00115, squid:S1488, squid:S1659

georgekankava:
squid:S00115 - Constant names should comply with a naming convention.
squid:S1488 - Local Variables should not be declared and then immediately returned or thrown.
squid:S1659 - Multiple variables should not be declared on the same line.
This pull request removes 40 minutes of technical debt.
https://dev.eclipse.org/sonar/rules/show/squid:S00115
https://dev.eclipse.org/sonar/rules/show/squid:S1488
https://dev.eclipse.org/sonar/rules/show/squid:S1659
George Kankava

### DIFF
--- a/src/main/java/com/yahoo/sketches/frequencies/ItemsSketch.java
+++ b/src/main/java/com/yahoo/sketches/frequencies/ItemsSketch.java
@@ -235,8 +235,8 @@ public class ItemsSketch<T> {
     final int type = extractFreqSketchType(pre0);       //Byte 6
 
     // Checks
-    final boolean preLongsEq1 = (preLongs == 1);        //Byte 0
-    final boolean preLongsEqMax = (preLongs == maxPreLongs);
+    final boolean preLongsEq1 = preLongs == 1;        //Byte 0
+    final boolean preLongsEqMax = preLongs == maxPreLongs;
     if (!preLongsEq1 && !preLongsEqMax) {
       throw new IllegalArgumentException(
           "Possible Corruption: PreLongs must be 1 or " + maxPreLongs + ": " + preLongs);
@@ -295,7 +295,8 @@ public class ItemsSketch<T> {
    */
   @SuppressWarnings("null")
   public byte[] serializeToByteArray(final ArrayOfItemsSerDe<T> serDe) {
-    final int preLongs, outBytes;
+    final int preLongs;
+    final int outBytes;
     final boolean empty = isEmpty();
     final int activeItems = getNumActiveItems();
     byte[] bytes = null;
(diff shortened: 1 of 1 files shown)


src/main/java/com/yahoo/sketches/frequencies/ItemsSketch.java
```diff
@@ -235,8 +235,8 @@ public ItemsSketch(final int maxMapSize) {
     final int type = extractFreqSketchType(pre0);       //Byte 6
 
     // Checks
-    final boolean preLongsEq1 = (preLongs == 1);        //Byte 0
-    final boolean preLongsEqMax = (preLongs == maxPreLongs);
+    final boolean preLongsEq1 = preLongs == 1;        //Byte 0
```
leerho: Technically correct, but harder to read.  No.

src/main/java/com/yahoo/sketches/frequencies/ItemsSketch.java
```diff
@@ -517,6 +526,47 @@ public int compareTo(final Row<T> that) {
     }
   }
 
+  @Override
```
leerho: This is wrong, but the reason is quite subtle.  Equals() and hashCode() are implemented when it is desirable to create hashTables of these objects or attempt to sort them.   But sketches are stochastic not deterministic processes.  This means that even if you feed the same identical stream of inputs into two different sketches of the same class, configured the same way, there is no guarantee that the internal contents of the two sketches will be the same (there are some exceptions, but it is better not to expect equality).  This is because random noise is intentionally inserted into sketch algorithms and is key to how they work.  Thus, it is not possible to construct an algorithm that can guarantee equality.  They may be similar, but not equal (but what good is that?).  Thus overriding equals() and hashCode() makes no sense.  That means creating HashTables or attempting to sort sketches also makes no sense.  The Identity case, where A.equals(A) will work and the Object's equals() and hashCode() already handle that just fine, but it is not very useful.  :)

src/main/java/com/yahoo/sketches/frequencies/ItemsSketch.java
```diff
@@ -550,9 +600,7 @@ public int compare(final Row<T> r1, final Row<T> r2) {
       }
     });
     
-    @SuppressWarnings("unchecked")
```
leerho: Eliminating this unchecked warning is not a good idea.  There is no way to get rid of the warning, so it just creates noise. It can either be here or at the method level, but it needs to stay.

src/main/java/com/yahoo/sketches/frequencies/ItemsSketch.java
```diff
@@ -295,7 +295,8 @@ public ItemsSketch(final int maxMapSize) {
    */
   @SuppressWarnings("null")
   public byte[] serializeToByteArray(final ArrayOfItemsSerDe<T> serDe) {
-    final int preLongs, outBytes;
+    final int preLongs;
```
leerho: I did implement many of the other changes in master.

PR #71, opened 2016-06-24
General quality fix - squid:S1226, squid:S1854, squid:S2184, squid:UselessParenthesesCheck

georgekankava:
squid:S1226 - Method parameters, caught exceptions and foreach variables should not be reassigned.
squid:S1854 - Dead stores should be removed.
squid:S2184 - Math operands should be cast before assignment.
This pull request removes 116 minutes of technical debt.
https://dev.eclipse.org/sonar/rules/show/squid:S1226
https://dev.eclipse.org/sonar/rules/show/squid:S1854
https://dev.eclipse.org/sonar/rules/show/squid:S2184
George Kankava

### DIFF
--- a/src/main/java/com/yahoo/sketches/hash/MurmurHash3.java
+++ b/src/main/java/com/yahoo/sketches/hash/MurmurHash3.java
@@ -76,7 +76,7 @@ public final class MurmurHash3 implements Serializable {
     // Get the tail
     long k1 = (rem == 0) ? 0 : key[tail]; //k2 -> 0
     // Mix the tail into the hash and return
-    return hashState.finalMix128(k1, 0, longs * 8); //convert to bytes
+    return hashState.finalMix128(k1, 0, longs * 8L); //convert to bytes
   }
   
   //--Hash of int[]----------------------------------------------------
@@ -107,7 +107,7 @@ public final class MurmurHash3 implements Serializable {
     int rem = ints - tail; // remainder ints: 0,1,2,3
     
     // Get the tail
-    long k1 = 0;
+    long k1;
     long k2 = 0;
     if (rem > 2) { //k1 -> whole; k2 -> partial
       k1 = getLong(key, tail, 2);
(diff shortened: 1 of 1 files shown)


src/main/java/com/yahoo/sketches/hash/MurmurHash3.java
```diff
@@ -256,12 +256,12 @@ void blockMix128(long k1, long k2) {
      * @return mix
      */
     private static long finalMix64(long h) {
-      h ^= h >>> 33;
-      h *= 0xff51afd7ed558ccdL;
-      h ^= h >>> 33;
-      h *= 0xc4ceb9fe1a85ec53L;
-      h ^= h >>> 33;
-      return h;
+      long result = h ^ h >>> 33;
```
leerho: It is not necessary to create an additional local variable on the stack!  In fact it would use more stack space and waste a few CPU cycles.  It does not make it any easier to read, so why do this?  Rejected.

src/main/java/com/yahoo/sketches/hash/MurmurHash3.java
```diff
@@ -76,7 +76,7 @@ private MurmurHash3() {}
     // Get the tail
     long k1 = (rem == 0) ? 0 : key[tail]; //k2 -> 0
     // Mix the tail into the hash and return
-    return hashState.finalMix128(k1, 0, longs * 8); //convert to bytes
+    return hashState.finalMix128(k1, 0, longs * 8L); //convert to bytes
```
leerho: It is even better to use Long.BYTES here instead of 8L.  Already changed in master.  Similar issues below.

src/main/java/com/yahoo/sketches/hash/MurmurHash3.java
```diff
@@ -107,7 +107,7 @@ private MurmurHash3() {}
     int rem = ints - tail; // remainder ints: 0,1,2,3
     
     // Get the tail
-    long k1 = 0;
+    long k1;
```
leerho: Well, you got it half right.  Both k1 and k2 should be uninitialized here, and then k2 = 0 added to the else clause.  This eliminates any unnecessary assignment.  Fixed in master.

PR #73, opened 2016-06-24
Code quality fix - squid:S00112, squid:S00122, squid:S1168, squid:S1213, squid:S2259

georgekankava:
squid:S00112 - Generic exceptions should never be thrown.
squid:S1168 - Empty arrays and collections should be returned instead of null.
squid:S1213 - The members of an interface declaration or class should appear in a pre-defined order.
squid:S2259 - Null pointers should not be dereferenced.
This pull request removes 91 minutes of technical debt.
https://dev.eclipse.org/sonar/rules/show/squid:S00112
https://dev.eclipse.org/sonar/rules/show/squid:S1168
https://dev.eclipse.org/sonar/rules/show/squid:S1213
https://dev.eclipse.org/sonar/rules/show/squid:S2259
George Kankava

### DIFF
--- a/src/main/java/com/yahoo/sketches/SketchesException.java
+++ b/src/main/java/com/yahoo/sketches/SketchesException.java
@@ -1,0 +1,26 @@
+package com.yahoo.sketches;
+
+/**
+ * Exception class for the library
+ */
+public class SketchesException extends RuntimeException {
+
+    public SketchesException() {
+    }
+
+    public SketchesException(String message) {
+        super(message);
+    }
+
+    public SketchesException(String message, Throwable cause) {
+        super(message, cause);
+    }
+
+    public SketchesException(Throwable cause) {
+        super(cause);
+    }
+
+    public SketchesException(String message, Throwable cause, boolean enableSuppression, boolean writableStackTrace) {
+        super(message, cause, enableSuppression, writableStackTrace);
+    }
+}
--- a/src/main/java/com/yahoo/sketches/tuple/CompactSketch.java
+++ b/src/main/java/com/yahoo/sketches/tuple/CompactSketch.java
@@ -8,6 +8,7 @@ import java.lang.reflect.Array;
 import java.nio.ByteOrder;
 
 import com.yahoo.sketches.Family;
+import com.yahoo.sketches.SketchesException;
 import com.yahoo.sketches.memory.Memory;
 import com.yahoo.sketches.memory.NativeMemory;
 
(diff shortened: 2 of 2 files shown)


src/main/java/com/yahoo/sketches/tuple/CompactSketch.java
```diff
@@ -96,15 +116,6 @@ public int getRetainedEntries() {
...
   }
 
-  // Layout of first 8 bytes:
-  // <pre>
-  // Long || Start Byte Adr:
-  // Adr: 
-  //      ||    7   |    6   |    5   |    4   |    3   |    2   |    1   |     0              |
-  //  0   ||                          |  Flags | SkType | FamID  | SerVer |  Preamble_Longs    |
-
-  private static final byte PREAMBLE_LONGS = 1;
```
leerho: This has been moved to the parent class as protected.

src/main/java/com/yahoo/sketches/tuple/CompactSketch.java
```diff
@@ -96,15 +116,6 @@ public int getRetainedEntries() {
     return keys_ == null ? 0 : keys_.length;
   }
 
-  // Layout of first 8 bytes:
```
leerho: What the review missed is that this code comment belongs to the toByteArray(), not to the PREAMBLE_LONGS constant.  Change rejected.

src/main/java/com/yahoo/sketches/tuple/CompactSketch.java
```diff
@@ -23,7 +24,16 @@
  */
 public class CompactSketch<S extends Summary> extends Sketch<S> {
 
-  static final byte serialVersionUID = 1;
```
leerho: Already changed in master to private

src/main/java/com/yahoo/sketches/tuple/CompactSketch.java
```diff
@@ -23,7 +24,16 @@
...
+  private static final byte serialVersionUID = 1;
+
```
leerho: This is in the wrong place.  See below.

src/main/java/com/yahoo/sketches/SketchesException.java
```diff
@@ -0,0 +1,26 @@
+package com.yahoo.sketches;
```
leerho: I actually like this suggestion.  However, to do this right would require creating several "SketchesException" sub-classes  and then redo all the thrown exceptions throughout the entire library.  I have implemented a simple model in master, but fully implementing this will take time.  I also don't feel we need to implement all the constructors when we only use one as this creates additional unused code that needs to be tested.  Rejected in this form. Implemented something simpler but better.

PR #74, opened 2016-06-24
Code quality fix - squid:S2178, squid:S00116, squid:S00122, squid:S1066, squid:UselessParenthesesCheck

georgekankava:
squid:S2178 - Short-circuit logic should be used in boolean contexts.
squid:S00116 - Field names should comply with a naming convention.
squid:S1066 - Collapsible "if" statements should be merged.
This pull request removes 61 minutes of technical debt.
https://dev.eclipse.org/sonar/rules/show/squid:S2178
https://dev.eclipse.org/sonar/rules/show/squid:S00116
https://dev.eclipse.org/sonar/rules/show/squid:S1066
George Kankava

### DIFF
--- a/src/main/java/com/yahoo/sketches/theta/DirectIntersection.java
+++ b/src/main/java/com/yahoo/sketches/theta/DirectIntersection.java
@@ -20,15 +20,15 @@ import com.yahoo.sketches.Util;
  * @author Kevin Lang
  */
 class DirectIntersection extends SetOperation implements Intersection {
-  private final short seedHash_;
+  private final short seedHash;
   //Note: Intersection does not use lgNomLongs or k, per se.
-  private int lgArrLongs_; //current size of hash table
-  private int curCount_; //curCount of HT, if < 0 means Universal Set (US) is true
-  private long thetaLong_;
-  private boolean empty_;
+  private int lgArrLongs; //current size of hash table
+  private int curCount; //curCount of HT, if < 0 means Universal Set (US) is true
+  private long thetaLong;
+  private boolean empty;
   
-  private final int maxLgArrLongs_; //max size of hash table
-  private final Memory mem_;
+  private final int maxLgArrLongs; //max size of hash table
+  private final Memory mem;
   
   /**
    * Construct a new Intersection target direct to the given destination Memory.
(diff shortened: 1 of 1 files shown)


src/main/java/com/yahoo/sketches/theta/DirectIntersection.java
```diff
@@ -88,44 +88,44 @@
...
+    curCount = srcMem.getInt(RETAINED_ENTRIES_INT);
+    thetaLong = srcMem.getLong(THETA_LONG);
     
-    if (empty_) {
-      if (curCount_ != 0) {
-        throw new IllegalArgumentException(
-            "srcMem empty state inconsistent with curCount: "+empty_+","+curCount_);
-      }
-      //empty = true AND curCount_ = 0: OK
+    if (empty && curCount != 0) {
```
leerho: This is the only change in this entire file that potentially has merit.  However, what was missed is the value of the comments embedded in the compound if/then/else clause:  The original code plus comments was specifically structured to help the reader understand subtleties in the code flow.  The comments were strategically placed in the position of the else clauses even though the else clauses are absent.  The suggested change ignores the value of the comments and in the suggested form the internal comment is now incorrect and would be confusing to the reader!  The compiler and JIT optimizer would restructure this compound statement anyway so there is NO performance impact on this change.  This change is rejected.

src/main/java/com/yahoo/sketches/theta/DirectIntersection.java
```diff
@@ -20,15 +20,15 @@
  * @author Kevin Lang
  */
 class DirectIntersection extends SetOperation implements Intersection {
-  private final short seedHash_;
```
leerho: I realize that the Google style guides and the Java religious don't like underscores.  Well, that is their opinion based on someone's concept of purity.  Tough!   I use underscores to clearly identify class members vs. local variables.  This avoids accidental name hiding and to avoid having to use _this._, which adds a lot of noise to the code.  All these underscore changes are rejected.

georgekankava: @leerho Yes it is actually up to your decision.

leerho:
I don't want to depreciate the work you have done.  But it would be more valuable if you focused on code problems that could cause true errors or poor performance and not focus on the minor style issues.  Your catching the "|" vs "||" was valuable as there is truly a performance gain to using "||". Thanks! 

Where it was clear that you misunderstood the code, I did try to make some changes to make it more clear and I have made comments in the PR.  This is a continuous, evolving process :)

Thanks again!

PR #81, opened 2016-07-23
Sketch now implements Serializable

thenobody:
Enabling serialisation for Sketch (and child classes) in order to facilitate instance transmission in Apache Spark.

### DIFF
--- a/src/main/java/com/yahoo/sketches/theta/Sketch.java
+++ b/src/main/java/com/yahoo/sketches/theta/Sketch.java
@@ -27,13 +27,15 @@ import com.yahoo.sketches.Family;
 import com.yahoo.sketches.SketchesArgumentException;
 import com.yahoo.sketches.memory.Memory;
 
+import java.io.Serializable;
+
 /**
  * The top-level class for all sketches. This class is never constructed directly. 
  * Use the UpdateSketch.builder() methods to create UpdateSketches.
  * 
  * @author Lee Rhodes 
  */
-public abstract class Sketch {
+public abstract class Sketch implements Serializable {
 
   static final int DEFAULT_LG_RESIZE_FACTOR = 3;   //Unique to Heap
   


src/main/java/com/yahoo/sketches/theta/Sketch.java
```diff
@@ -27,13 +27,15 @@
...
+import java.io.Serializable;
+
 /**
  * The top-level class for all sketches. This class is never constructed directly. 
  * Use the UpdateSketch.builder() methods to create UpdateSketches.
  * 
  * @author Lee Rhodes 
  */
-public abstract class Sketch {
+public abstract class Sketch implements Serializable {
```
leerho: Thank you for your contribution!  
However, I have specifically avoided Java default serialization because it is so horribly inefficient. In a large system environment with literally billions of sketches this would reduce performance considerably.  Also,  Java default serialization will not work at all for the Direct* sketches, which are children of Sketch, since the data for these sketches is off the Java heap.  The preferred SerDe strategy is to use the toByteArray() methods and the Sketches methods heapify*() and wrap*(), which take a Memory object that wraps a byte array either on or off the Java heap.  Some big data systems, such as Hadoop/Pig provide mechanisms for efficiently transporting binary data.  For those that don't, It is much more efficient to encode the above binaries using Base64.  This can be done outside the context of the sketches themselves.  Because these transport capabilities vary across systems, the sketch library does the serialization and deserialization into these binary objects, the rest is up to the Systems Engineers to decide the most efficient way to do this.

thenobody:
Thanks for a quick response.
I'm absolutely with you regarding the performance point. However, I did actually spend quite a bit of time before trying to do get around the serialisation issue using, as you mentioned, the toByteArray logic (using Kryo) to no avail. Unfortunately, Spark absolutely _requires_ Serializable classes when used in the distributed context. This is because even though the Sketch class itself is made serialisable by Kryo Spark still needs to distribute the code/closures (which include references to the Sketches) over the wire. And this is unfortunately done using the regular Java serialisation (I've tried to override this behaviour but no luck, Spark does not support anything else). The only way I was able to get past the code serialisation was to introduce Serializable to the top-level Sketch class.
Given this, would you maybe reconsider enabling Serializable for Sketch? Or am I possibly missing something Spark-wise?

thenobody:
Just a small update...
I was able to get around the closure serialisation by wrapping the Sketch in an `Option` (in my case it's `aggregateByKey[Option[Sketch]](None) ((acc: Option[Sketch], value: String => ...), ...)` instead of `aggregateByKey[Sketch](Sketches.updateSketchBuilder().build) ((acc: Sketch, value: String => ...), ...)`) and then handling the initialisation of the sketch inside the closure that handles individual iterations.
This works but I don't like the fact that it introduces wasteful `Some` initialisation over and over again for every single value and the necessity of checking if the accumulator has been initialised in every step. Also, it feels like a lot like hackery, especially when reducing two counters into a single one, i.e.
```
(counter1: Option[Sketch], counter2: Option[Sketch]) => {
  val union = Sketches.setOperationBuilder().buildUnion
  union.update(counter1.get)
  union.update(counter2.get)
  Some(union.getResult)
}
```

Oh and outside of this, the `Serializable` also comes in handy when persisting the counters using `saveAsObjectFile` 😄 I know I could transform them manually using `toByteArray` but it brings in boilerplate for both writing and reading the output.

leerho:
One of the fundamental problems, which I mentioned in my first reply, is that implementing Serializable in Sketch, which is the root class of all the Theta Sketches, might work for the heap-based sketches but will certainly fail for the off-heap sketches.   

Sketches are _streaming algorithms_ and are _stateful_ by design. Attempting to force them into a stateless paradigm will result in orders-of-magnitude poorer performance.   It is like pounding a square peg into a round hole.   I think that you realize that the absolutely worst thing to do would be (similar to what you mentioned):

for (each value) {

leerho:
BTW, DataSketches have also been tightly integrated into Druid and into Storm, so I know it can be done.

leerho:
@thenobody

Another Spark user asked a similar question on https://groups.google.com/forum/#!forum/sketches-user and we suggested using a wrapper class.  You can view this conversation under "Usable in Spark?" by Prasad Chalasani on Jun 23rd. You could also try that.

thenobody:
@leerho 
Thanks for the follow-up, much appreciated :)
The wrapper approach is quite similar to what I implemented using the `Option` + Kryo serialisation combo (although, the wrapper would also make `saveAsObjectFile` work as well). I might just as well go with the wrapper approach since it promotes considerably less bolierplate/hackery than manually (de)serialising the sketches between Spark jobs and stages.

Regarding the on/off heap duality, good point, making `Sketch` implement `Serializable` would just lead to trouble. I guess one could then try to argue to make only the `Heap` family of classes implement `Serializable` but that doesn't seem like a good idea (I would say it'd make the API convoluted).

Regarding the statefulness of the sketch, I absolutely agree that making the union a stateless pure function would be a waste of resources (apologies but I don't see how serialisation influences this).
Just to give a bit of a context, the point in code where I'd need to perform the union actually expects the sketch to be mutable. The processing flow is as follows:
```
val partitions: Iterable[Seq[String]] = ... // partitions of data points
val partialSketches = partitions.map { partition =>
  partition.foldLeft(EmptySketch()) ((sketch, dataPoint) =>
    sketch.update(datePoint)
    sketch
  )
}
val finalSketch = partialSketches.reduce((leftSketch, rightSketch) => ... // perform and return a union)
```
Each operation on a partition happens as a separate Spark step.
Since the Sketch itself doesn't support merging another Sketch into it I need to use the UnionBuilder for the final reduce. Originally, I was hoping to utilise a mechanism similar to how HyperLogLogs can be [merged here](https://github.com/addthis/stream-lib/blob/master/src/main/java/com/clearspring/analytics/stream/cardinality/HyperLogLog.java#L218) (modifies the state of the _'left'_ counter by merging in the _'right'_ counter).

On a bit more technical side note, the place where the serialisation comes into play here is when Spark sends partially updated sketches (the result of the `foldLeft`) to the reducers to merge them into a single final result. There are other approaches how to achieve this same result (a single final counter) but this one seems to be the most memory/resource efficient (from Spark's point of view that is).

To sum up, I think I can make this work even without 'shoehorning' `Serializable` onto the `Sketch` itself using the wrapper approach you mentioned. Would it make sense though to maybe consider adding a `Sketch.merge(Sketch other)` that would modify the original Sketch? As far as I understand, the KMV algo supports this in a quite straightforward way.

leerho:
@thenobody

I would like to copy this discussion into the sketch-users forum so that other Spark users can more easily find it.

[https://groups.google.com/forum/#!newtopic/sketches-user](https://groups.google.com/forum/#!newtopic/sketches-user)

thenobody:
@leerho by all means :) please feel free

PR #83, opened 2016-07-27
refactor duplicated code

pdeva:


leerho:
Thanks for this!  I pulled this into a different branch with the same name for closer examination and to be able to run my Checkstyle and FindBugs, which are not integrated into pom yet :) 

This also allows me to clean up the similar logic in the ItemsSketch.

PR #98, opened 2016-09-07
Fix unit tests to actually free memory after catching exception

praveev:


leerho:
Thank you for your contribution!  But I rewrote these to use the try -- finally clause instead.

PR #116, opened 2016-11-17
NativeMemory: Respect arrayOffset for onheap ByteBuffers.

gianm:
Previously, the test `testSliceHeapByteBuffer` would fail by returning 1 instead of 2. `testSliceDirectByteBuffer` passed even before this patch, but is offered for completeness.


leerho:
Really good suggestion, thanks!

PR #150, opened 2017-06-03
remove modules, shift sketches/src and sketches/test to be off base d…

jmalkin:
…irectory


jmalkin:
@AlexanderSaydakov Please review, just in case I missed something.

jmalkin:
i know the issue.. gotta figure out how to safely adjust the test bound so that it's testing something meaningful but with a significantly lower probability of failure

PR #164, opened 2017-10-12
Fix bug in HeapUpdatesketch.toByteArray() that didn't set ...

jmalkin:
... resize factor. Could cause infinite loop if wrapping an under-full serialized sketch and updating it.


AlexanderSaydakov:
looks good to me

PR #169, opened 2017-11-26
Singleton theta

leerho:
I was suspicious that the SingleItemSketch was still not completely implemented.  So I created tests that checked all the different set operations and on-and off heap.  Sure enough, there were some gaps.  Any circumstance where a CompactSketch ends up with only one entry should result in a SingleItemSketch.  This now works.  Along the way I found a bug with the AlphaSketch where it wasn't able to heapify.  This is also fixed.  Also the initialization of the AnotB operation was not quite correct so I fixed that.   I also added some more unit tests for the Theta package where I found opportunities to do so.   Most users would never see these new discrepancies I found.


jmalkin:
I didn't try to understand the diffs on the tests, but the main code seems ok. Merge away!

PR #183, opened 2018-02-03
specialized getRank for speed, avoid one array allocation in PMF and CDF

AlexanderSaydakov:


src/main/java/com/yahoo/sketches/quantiles/DoublesPmfCdfImpl.java
```diff
@@ -93,7 +86,7 @@
   static void bilinearTimeIncrementHistogramCounters(final DoublesBufferAccessor samples,
```
leerho: @AlexanderSaydakov 
Style comment: except for here and the method just below, we have samples, weight, splitPoints, counters all in a single line.  I suggest we be consistent with a line-return before the first parameter and all the params on the same line.

PR #186, opened 2018-03-12
Respect comparator in items sketch

navis:
comparator handed over to items sketch is not used when sorting, making exception "Values must be unique, monotonically increasing and not NaN."


yahoocla:
Thank you for submitting this pull request, however I do not see a valid CLA on file for you. Before we can merge this request please visit https://yahoocla.herokuapp.com/ and agree to the terms. Thanks! :smile:

navis:
signed CLA

leerho:
Thanks for finding this.  We will look closely at your proposed solution. :)

leerho:
Thank you very much for finding this.  We decided to fix it a little differently, which doesn't impact performance, as well as made sure we had tests that covered both cases.

PR #189, opened 2018-03-15
Support Filter in tuple sketches

inigoillan:
This PR adds support for filtering tuple sketches values based on summary

### DIFF
--- a/src/main/java/com/yahoo/sketches/tuple/Filter.java
+++ b/src/main/java/com/yahoo/sketches/tuple/Filter.java
@@ -9,18 +9,15 @@ import java.util.function.Predicate;
  */
 public class Filter<T extends Summary> {
     private Predicate<T> predicate;
-    private SummaryFactory<T> summaryFactory;
 
     /**
-     * Filter constructor with a {@link Predicate} and {@link SummaryFactory}
-     *
-     * @param predicate Predicate to use in this filter. If the Predicate returns False, the element is discarded.
+     * Filter constructor with a {@link Predicate}
+     *  @param predicate Predicate to use in this filter. If the Predicate returns False, the element is discarded.
      *                 If the Predicate returns True, then the element is kept in the {@link Sketch}
-     * @param summaryFactory Summary Factory to use to build a new summary
+     *
      */
-    public Filter(Predicate<T> predicate, SummaryFactory<T> summaryFactory) {
+    public Filter(Predicate<T> predicate) {
         this.predicate = predicate;
-        this.summaryFactory = summaryFactory;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jmalkin:
There's an issue here in that the result isn't really a valid sketch. The estimated count from the full sketch scaled by the % that match your predicate is the best estimate of the total number from the full stream matching the predicate. But removing items like this will make errors in subsequent uses worse in ways that mean the error bounds are unknown.

It'd be fine to return a cardinality estimate based on applying the predicate.

inigoillan:
Hi Jon,

Thanks for your comment. As I understand, the error bounds on the Tuple sketches depend on the theta value and the number of retained entries. While it's true the theta value depends directly by the number of elements added to the sketch, I still don't see a reason why those are still unknown. 

I am wondering as well, how this actually would work for the intersection. If you think a bit about it, a filter is just an intersection between 2 sets, one of it declared by enumeration and the other one declared by comprehension. Indeed, the approach followed by the [Intersection class](https://github.com/DataSketches/sketches-core/blob/master/src/main/java/com/yahoo/sketches/tuple/Intersection.java) looks quite similar to this one.

jmalkin:
We debated this internally since we all had the same initial reaction, and we decided that yes, it is a valid thing to do. It will indeed function as an intersection, meaning that there is some accuracy degradation but with known bounds. I modified my initial comment to note the retraction.

jmalkin:
The other thing we discussed is whether we want this as a stand-alone class or if it makes more sense to have something an applyPredicate(Predicate p) method on the sketch itself. So we'd appreciate hearing why you went with this route? The are pros and cons to both approaches, of course, so it's not like there's a single correct way.

inigoillan:
Thanks for the clarification.

The reasons why I choose this approach are 2:

- The main one is it follows a similar pattern to the `Filter` and `Intersection ` classes, so it feels consistent with the rest of the code.
- The second one is the semantics are not very clear. i.e: for the UpdatableSketch whether this operation is immutable and if the predicate is also applied to the original sketch.

As well, implementing this as a method in the Sketch<S> class would require a circular dependency on the QuickSelectSketch and introduce a dependency of the Factory method (or rather have several implementations, as CompactSketch and QuickSelectSketch only share Sketch<S> as parent)

All in all, I'm open to changes in the code

inigoillan:
@jmalkin Is it possible to move this forward?

jmalkin:
Sorry, I thought I'd submitted the comments already. Apparently they've been sitting there for a while now.

inigoillan:
I addressed your comments. Could you have a look? @jmalkin

jmalkin:
Sorry, had only phone-based network access recently. Will try to get to this shortly.

PR #190, opened 2018-03-20
Use thread local random instead of global, shared one

navis:
Sometimes it's required to get determinitic result from sketches. But currently random instance in ItemsSketch is static one and hard to provide seed for that purpose in multi-threaded env. And also, java doc in Random class sugests to use thread-local to avoid contention.

### DIFF
--- a/src/main/java/com/yahoo/sketches/quantiles/ItemsMergeImpl.java
+++ b/src/main/java/com/yahoo/sketches/quantiles/ItemsMergeImpl.java
@@ -191,7 +191,7 @@ final class ItemsMergeImpl {
       final T[] bufC, final int startC, // output
       final int kC, // number of items that should be in the output
       final int stride) {
-    final int randomOffset = ItemsSketch.rand.nextInt(stride);
+    final int randomOffset = ItemsSketch.rand.get().nextInt(stride);
     final int limC = startC + kC;
     for (int a = startSrc + randomOffset, c = startC; c < limC; a += stride, c++ ) {
       bufC[c] = bufSrc[a];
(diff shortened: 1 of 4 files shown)


jmalkin:
Can you elaborate a bit more on the multi-threaded environment usage? The library is not thread-safe in general.

navis:
I think you've heard of druid, which uses sketch library extensively. What I'm trying with sketch is optimizing query like "group by X", of which X is high-cardinality column. Cause group-by is very memory intensive work, I want to split the query into sequential queries on column X ("group-by x where x < 100" + "group-by x where 100 < x < 200" + "group-by x where x > 200", for example)

For that purpose, it's not needed to be exact histogram of X but it's a little efficient to return same result (histogram) from merged result of sketches from segments because filter evaluations (bitmap, in druid), which takes several hundreds of milliseconds to build, are cached and reused. If thresholds from histogram is ever changing, druid need to calculate this every time.  

I know sketch can be different from same stream of data but again cache in druid make it possible to return always the same sketch from a segment. So what we need is to set seed before starting merging of sketches which is executed by single thread to get deterministic result. And cause druid can handle multiple queries concurrently, it need to be thread-local or per instance base.

Could this be enough explanation?

AlexanderSaydakov:
I have big concerns regarding any attempts to defeat randomness in production settings. While it is, probably, a good idea to use thread local random generator, the proper way, as I see it, would be to use the specialized class ThreadLocalRandom. This class should provide performance benefits in a multi-threaded environment, and, perhaps, even in a single-threaded one, but it intentionally prohibits setting the seed.
In this pull request ThreadLocal<Random> is used instead of the specialized ThreadLocalRandom, to work around the prohibition. It seems to me that this approach won't have the performance advantage of ThreadLocalRandom, and also provide dangerous possibilities of correlated random generators in different threads.

leerho:
At some point we may look at moving to `ThreadLocalRandom` for better update performance, but it is not a high priority right now.  

It is not a good idea to make the random generator deterministic in production due to the possibilities of spurious correlations as Alex mentions above. (Very localized and controlled test environments can be an exception).  At some point in the future we may decide to disallow any deterministic seeding of the random generator.

This PR will now be closed.

PR #196, opened 2018-04-18
Fix SerDe Issue 195

leerho:


src/test/java/com/yahoo/sketches/quantiles/DoublesUnionImplTest.java
```diff
@@ -693,6 +693,15 @@ public void isSameResourceDirect() {
     Assert.assertFalse(union.isSameResource(mem2));
   }
 
+  @SuppressWarnings("unused")
+  @Test
+  public void checkSerDeIssue165() {
```
hpx7: did you mean 195?

AlexanderSaydakov:
The fix looks fine to me. Let's rename the test to either checkSerDeIssue195 or, perhaps, something like emptyUnionSerDe and assert a few things about the result.

hpx7:
Thanks for the quick turnaround! Could we tag a release for this?

leerho:
Hopefully soon.  We need a few days to review all the changes we have made
since the last release.

PR #197, opened 2018-04-18
Removed old EpsilonFromK model

leerho:


src/main/java/com/yahoo/sketches/kll/KllFloatsSketch.java
```diff
@@ -781,6 +783,17 @@ public static KllFloatsSketch heapify(final Memory mem) {
     return new KllFloatsSketch(mem);
   }
 
+  /**
+   * Checks the validity of the given value k
+   * @param k must be greater than 1 and less than 65536.
+   */
+  static void checkK(final int k) {
+    if ((k < MIN_K) || (k > MAX_K) || !isPowerOf2(k)) {
```
AlexanderSaydakov: should not be restricted to power of 2

PR #198, opened 2018-04-20
Move Heapify and wrap to base DoublesUnion class,

leerho:
deprecate original location.


leerho:
Added some unit tests, didn't realize you had already merged.

PR #202, opened 2018-05-23
concurrent theta sketch

Eshcar:
@leerho 
@AlexanderSaydakov 

Concurrent theta sketches branch:

The implementation is still missing proper documentation which I plan to add.
Also note that since the concurrent sketch is in a separate package and it extends UpdateSketch I changed the visibility of some methods from package visibility to public.
Another option is to add the concurrent sketches to the theta package (instead of concurrent.theta).

### DIFF
--- a/src/main/java/com/yahoo/sketches/concurrent/theta/ConcurrentUpdateSketch.java
+++ b/src/main/java/com/yahoo/sketches/concurrent/theta/ConcurrentUpdateSketch.java
@@ -1,0 +1,85 @@
+package com.yahoo.sketches.concurrent.theta;
+
+import com.yahoo.sketches.theta.Sketch;
+import com.yahoo.sketches.theta.UpdateSketch;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * @author eshcar
+ */
+public class ConcurrentUpdateSketch extends UpdateSketchComposition {
+
+  private static int PARALLELISM_LEVEL = 3;
+  protected static ExecutorService PROPAGATION_EXECUTOR_SERVICE = Executors.newWorkStealingPool
+      (PARALLELISM_LEVEL);
+
+  private volatile long theta_;
+  private volatile double estimation_;
+  // A flag to coordinate between several propagation threads
+  private AtomicBoolean propagationInProgress_;
+
+  // package visibility constructors - to be created only by factory
+  ConcurrentUpdateSketch(final UpdateSketch delegattee) {
+    super(delegattee);
+    theta_ = getThetaLong();
+    estimation_ = 0;
+    propagationInProgress_ = new AtomicBoolean(false);
+  }
+
+  public void propagate(final Sketch sketchIn, AtomicBoolean propagationInProgress) {
+    BackgroundThetaPropagation job = new BackgroundThetaPropagation(sketchIn,
+        propagationInProgress);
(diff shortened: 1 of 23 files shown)


src/main/java/com/yahoo/sketches/concurrent/theta/ConcurrentUpdateSketch.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * @author eshcar
+ */
+public class ConcurrentUpdateSketch extends UpdateSketchComposition {
+
+  private static int PARALLELISM_LEVEL = 3;
+  protected static ExecutorService PROPAGATION_EXECUTOR_SERVICE = Executors.newWorkStealingPool
```
ebortnik: Does the DataSketches package have its own executor pools?

src/main/java/com/yahoo/sketches/concurrent/theta/ConcurrentUpdateSketch.java
```diff
@@ -0,0 +1,85 @@
...
+    public BackgroundThetaPropagation(Sketch sketch, AtomicBoolean localThreadPropagationFlag) {
+      this.sketch_ = sketch;
+      this.localThreadPropagationFlag_ = localThreadPropagationFlag;
+    }
+
+    @Override public void run() {
+      assert getVolatileTheta() <= sketch_.getThetaLong();
+
+      while(!propagationInProgress_.compareAndSet(false,true)) {
+        //busy wait until can propagate
```
ebortnik: Dangerous, potential for livelock.  Can give up / throw exception after so many failures?

src/main/java/com/yahoo/sketches/concurrent/theta/UpdateSketchComposition.java
```diff
@@ -0,0 +1,107 @@
+package com.yahoo.sketches.concurrent.theta;
+
+import com.yahoo.memory.WritableMemory;
+import com.yahoo.sketches.Family;
+import com.yahoo.sketches.ResizeFactor;
+import com.yahoo.sketches.theta.UpdateReturnState;
+import com.yahoo.sketches.theta.UpdateSketch;
+
```
ebortnik: External comment please - why do we need this class

src/main/java/com/yahoo/sketches/concurrent/theta/ConcurrentUpdateSketch.java
Eshcar: Maybe there is, I'm not sure.
I couldn't find ExecutorService in the package.

src/main/java/com/yahoo/sketches/concurrent/theta/UpdateSketchComposition.java
Eshcar: I will add documentation.
Basically this class aims to eliminate code duplication. Currently 2 classes extend it and don't need to replicate the code in it.

src/main/java/com/yahoo/sketches/concurrent/theta/ConcurrentUpdateSketch.java
leerho: The DS package does not have executor pools because the entire package was designed to be single threaded... until now :)

Eshcar:
refactor moved new classes to the theta package so now changes to existing files are much smaller

PR #203, opened 2018-06-01
Fix ArrayOfDoublesUnion to be compatible with new Memory > 0.11.0.

leerho:


src/main/java/com/yahoo/sketches/tuple/ArrayOfDoublesUnion.java
```diff
@@ -190,16 +190,23 @@ void setThetaLong(final long theta) {
...
-      return isWritable ? new DirectArrayOfDoublesUnion(sketch, mem)
-          : new DirectArrayOfDoublesUnionR(sketch, mem);
+      if (isWritable) {
+        sketch = new DirectArrayOfDoublesQuickSelectSketch(mem, seed);
+        union = new DirectArrayOfDoublesUnion(sketch, mem);
+      } else {
+        sketch = new DirectArrayOfDoublesQuickSelectSketchR(mem, seed);
+        union = new DirectArrayOfDoublesUnionR(sketch, mem);
+      }
+      return union; //TODO set union theta?
```
leerho: I notice that you do not set union.theta_ here as you do on line 232.  Oversight?

src/main/java/com/yahoo/sketches/tuple/ArrayOfDoublesUnion.java
```diff
@@ -210,13 +217,18 @@ static ArrayOfDoublesUnion wrapUnionImpl(final WritableMemory mem, final long se
...
+    if (isWritable) {
+      final WritableMemory sketchMem = mem.writableRegion(ArrayOfDoublesUnion.PREAMBLE_SIZE_BYTES,
+          mem.getCapacity() - ArrayOfDoublesUnion.PREAMBLE_SIZE_BYTES);
+      sketch = new DirectArrayOfDoublesQuickSelectSketch(sketchMem, seed);
+      union = new DirectArrayOfDoublesUnion(sketch, mem);
+
+    } else {
+      final Memory sketchMem = mem.region(ArrayOfDoublesUnion.PREAMBLE_SIZE_BYTES,
+          mem.getCapacity() - ArrayOfDoublesUnion.PREAMBLE_SIZE_BYTES);
+      sketch = new DirectArrayOfDoublesQuickSelectSketchR((WritableMemory)sketchMem, seed);
```
leerho: Why does a read-only sketch require a WritableMemory input?

PR #209, opened 2018-07-09
Eshcar conc theta

leerho:
Moved your branch to a branch on sketches-core.

This is much cleaner!

You need to enable CheckStyle in your IDE and then point it to /tools/SketchesCheckstyle.xml

It should highlight public methods that are missing javadocs: namely: 
  ConcurrentThetaContext
  ConcurrentUpdateSketch

I already fixed a number of missing "final" modifiers.  

The current code has compile errors because of missing Logger.  However, I've decided to allow the single dependency: org.slf4j / slf4j-api.  This enables a user to pull in their own logger of choice at **runtime**, which can be the apache.commons.logging or the built in Java one or numerous other ones.  You can see how it is used in the Memory repository.  It adds virtually no overhead if you don't specify a logger at all.  Then when you attach your logger of choice, it pulls it in under a common interface.

### DIFF
--- a/src/main/java/com/yahoo/sketches/theta/ConcurrentThetaContext.java
+++ b/src/main/java/com/yahoo/sketches/theta/ConcurrentThetaContext.java
@@ -1,0 +1,65 @@
+package com.yahoo.sketches.theta;
+
+import com.yahoo.sketches.theta.CompactSketch;
+import com.yahoo.sketches.theta.ConcurrentUpdateSketch;
+import com.yahoo.sketches.theta.UpdateReturnState;
+import com.yahoo.sketches.theta.UpdateSketch;
+import com.yahoo.sketches.theta.UpdateSketchBuilder;
+import com.yahoo.sketches.theta.UpdateSketchComposition;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * @author eshcar
+ */
+public class ConcurrentThetaContext extends UpdateSketchComposition {
+
+  UpdateSketch localSketch_;
+  private AtomicBoolean propagationInProgress_;
+
+  ConcurrentThetaContext(ConcurrentUpdateSketch shared, UpdateSketch local) {
+    super(shared);
+    localSketch_ = local;
+    propagationInProgress_ = new AtomicBoolean(false);
+    localSketch_.setThetaLong(shared.getVolatileTheta());
+  }
+
+  ConcurrentThetaContext(ConcurrentUpdateSketch shared, UpdateSketchBuilder usb) {
+    super(shared);
+    localSketch_ = usb.build();
+    propagationInProgress_ = new AtomicBoolean(false);
+    localSketch_.setThetaLong(shared.getVolatileTheta());
+  }
+
+  public ConcurrentUpdateSketch getSharedSketch() {
+    return (ConcurrentUpdateSketch) delegatee_;
+  }
+
+
+  @Override
+  UpdateReturnState hashUpdate(final long hash) {
+    UpdateReturnState ret = localSketch_.hashUpdate(hash);
+    if (localSketch_.isOutOfSpace(localSketch_.getRetainedEntries(false)+1)
+        && localSketch_.getLgArrLongs() > localSketch_.getLgNomLongs()) {
+        // local theta is going to be changed upon next update
+        propagateToSharedSketch();
+    }
+    return ret;
+  }
+
+  @Override
+  public double getEstimate() {
+    return getSharedSketch().getEstimationSnapshot();
+  }
+
+  private void propagateToSharedSketch() {
+    while (propagationInProgress_.get()) {} //busy wait
+
+    CompactSketch compactSketch = localSketch_.compact();
+    propagationInProgress_.set(true);
+    getSharedSketch().propagate(compactSketch, propagationInProgress_);
+    localSketch_.reset();
+    localSketch_.setThetaLong(getSharedSketch().getVolatileTheta());
+  }
+
+}
--- a/src/main/java/com/yahoo/sketches/theta/ConcurrentUpdateSketch.java
+++ b/src/main/java/com/yahoo/sketches/theta/ConcurrentUpdateSketch.java
@@ -1,0 +1,86 @@
+package com.yahoo.sketches.theta;
+
+import com.yahoo.sketches.theta.Sketch;
+import com.yahoo.sketches.theta.UpdateSketch;
+import com.yahoo.sketches.theta.UpdateSketchComposition;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * @author eshcar
+ */
+public class ConcurrentUpdateSketch extends UpdateSketchComposition {
+
+  private static int PARALLELISM_LEVEL = 3;
+  protected static ExecutorService PROPAGATION_EXECUTOR_SERVICE = Executors.newWorkStealingPool
+      (PARALLELISM_LEVEL);
+
+  private volatile long theta_;
+  private volatile double estimation_;
+  // A flag to coordinate between several propagation threads
+  private AtomicBoolean propagationInProgress_;
+
+  // package visibility constructors - to be created only by factory
+  ConcurrentUpdateSketch(final UpdateSketch delegattee) {
+    super(delegattee);
+    theta_ = getThetaLong();
+    estimation_ = 0;
+    propagationInProgress_ = new AtomicBoolean(false);
+  }
+
+  public void propagate(final Sketch sketchIn, AtomicBoolean propagationInProgress) {
+    BackgroundThetaPropagation job = new BackgroundThetaPropagation(sketchIn,
+        propagationInProgress);
+    PROPAGATION_EXECUTOR_SERVICE.execute(job);
+  }
+
+  public double getEstimationSnapshot() {
+    return estimation_;
+  }
+
+  public long getVolatileTheta() {
+    return theta_;
+  }
+
+  private class BackgroundThetaPropagation implements Runnable {
+
+    private Sketch sketch_;
+    private AtomicBoolean localThreadPropagationFlag_;
+
+    public BackgroundThetaPropagation(Sketch sketch, AtomicBoolean localThreadPropagationFlag) {
+      this.sketch_ = sketch;
+      this.localThreadPropagationFlag_ = localThreadPropagationFlag;
+    }
+
+    @Override public void run() {
+      assert getVolatileTheta() <= sketch_.getThetaLong();
+
+      while(!propagationInProgress_.compareAndSet(false,true)) {
+        //busy wait until can propagate
+      }
+      //At this point we are sure only a single thread is propagating data to the shared sketch
+
+      // propagate values from input sketch one by one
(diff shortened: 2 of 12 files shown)


leerho:
Your new classes have zero code coverage.  This needs to be fixed.  

Also your classes have almost no javadocs.  For example, what is the ConcurrentThetaContext for?

It doesn't look like the TestPerformanceTheta runs with TestNG.  If this is a long running performance characterization test, we need to move / copy it under the Characterization repository.  This is where we do all of our long running characterization tests.  

For Unit Tests we need simple checks against all the code paths that they work properly.  You should  install Clover on your system (it is now open source) and run your coverage tests yourself.  I find it more comprehensive than the online coverage tools.

I will add more comments as I get deeper into the code.

src/test/java/com/yahoo/sketches/theta/TestPerformanceTheta.java
```diff
@@ -0,0 +1,245 @@
...
+          + "CONCURRENT/LOCK_BASED/BASELINE";
+      LOG.info(msg);
+      throw new RuntimeException(msg);
+    }
+
+    for (long i = 0; i < 10000000; i++) {
+      sketchToInit.update(i);
+      if((i%100000)==0){
+        System.out.print(".");
+      }
```
leerho: Other that drawing a line of dots, what is the purpose of this? 
Why not print to LOG rather than to System.out ?

src/test/java/com/yahoo/sketches/theta/TestPerformanceTheta.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+    System.out.println();
```
leerho: Why not print to LOG rather than to System.out ?

src/main/java/com/yahoo/sketches/theta/UpdateSketchComposition.java
```diff
@@ -0,0 +1,105 @@
+package com.yahoo.sketches.theta;
+
+import com.yahoo.memory.WritableMemory;
+import com.yahoo.sketches.Family;
+import com.yahoo.sketches.ResizeFactor;
+
+/**
+ * @author eshcar
+ */
+public class UpdateSketchComposition extends UpdateSketch {
```
leerho: This indirection is unfortunate as this will likely be a performance hit on a single thread 1:1 comparison with the non-concurrent sketch. You make it up with parallel threads of course.

Nonetheless, we need some documentation to give guidance what the trade-offs are so users don't unwittingly always throw in the concurrent version when it is not needed.

src/main/java/com/yahoo/sketches/theta/ConcurrentUpdateSketch.java
```diff
@@ -0,0 +1,83 @@
...
+      }
+      //At this point we are sure only a single thread is propagating data to the shared sketch
+
+      // propagate values from input sketch one by one
+      final long[] cacheIn = sketch_.getCache();
+      final int numEntries = sketch_.getRetainedEntries(false);
+      for (int i = 0; i < numEntries; i++) {
+        // pre-condition: cacheIn values are sorted in ascending order
+        final long hashIn = cacheIn[i];
+        if (hashIn >= getVolatileTheta()) {
```
leerho: Since we know we are operating in a single thread here, why don't we keep the volatile theta in a local long and then store it back into the volatile at the end of the propagation?

Also, I don't see a number of checks that are performed in the UnionImpl merge code:

1) if the sketch is off-heap, you need to access the values via Memory.

2) The UnionImpl code performs the "Theta rule" before and after each sketch merge.  I don't see it here at all.  I don't think you can assume that the lowest theta is always the input sketch (or the shared sketch).

3) You are compacting each input sketch prior to your merge, (In order to leverage the early stop, I presume) But it is not at all clear that this is a good performance trade-off.  Have you characterized this?

***
Why are you updating the estimate after every merge ?

src/main/java/com/yahoo/sketches/theta/ConcurrentUpdateSketch.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * @author eshcar
+ */
+public class ConcurrentUpdateSketch extends UpdateSketchComposition {
+
+  private static int PARALLELISM_LEVEL = 3;
```
leerho: I would be nice if you explained in a code comment what PARALLELISM_LEVEL = 3 means and why 3 was chosen.

src/main/java/com/yahoo/sketches/theta/ConcurrentUpdateSketch.java
```diff
@@ -0,0 +1,83 @@
...
+
+  // package visibility constructors - to be created only by factory
+  ConcurrentUpdateSketch(final UpdateSketch delegattee) {
+    super(delegattee);
+    theta_ = getThetaLong();
+    estimation_ = 0;
+    propagationInProgress_ = new AtomicBoolean(false);
+  }
+
+  public void propagate(final Sketch sketchIn, final AtomicBoolean propagationInProgress) {
```
leerho: Why does this need to be public?

src/main/java/com/yahoo/sketches/theta/ConcurrentUpdateSketch.java
```diff
@@ -0,0 +1,83 @@
...
+    final BackgroundThetaPropagation job = new BackgroundThetaPropagation(sketchIn,
+        propagationInProgress);
+    PROPAGATION_EXECUTOR_SERVICE.execute(job);
+  }
+
+  public double getEstimationSnapshot() {
+    return estimation_;
+  }
+
+  public long getVolatileTheta() {
```
leerho: public ?

src/main/java/com/yahoo/sketches/theta/ConcurrentThetaContext.java
```diff
@@ -0,0 +1,58 @@
+package com.yahoo.sketches.theta;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * @author eshcar
+ */
+public class ConcurrentThetaContext extends UpdateSketchComposition {
```
leerho: public ?

Eshcar: Yes, this is the channel through which threads update the sketch - the application needs a context per thread.

src/main/java/com/yahoo/sketches/theta/ConcurrentUpdateSketch.java
Eshcar: I agree.
Basically this is the number of threads allocated to run propagation from local sketches to shared sketches in the background.
3 is not carved in stone, maybe it should be configurable, but we should be able to set a default value based on benchmarks.

src/main/java/com/yahoo/sketches/theta/ConcurrentUpdateSketch.java
Eshcar: I'll check, perhaps package visibility is enough

src/main/java/com/yahoo/sketches/theta/ConcurrentUpdateSketch.java
Eshcar: likewise

src/main/java/com/yahoo/sketches/theta/ConcurrentUpdateSketch.java
Eshcar: I will check all the points you raise here .
Not sure what you mean by Theta rule, but we maintain an invariant  - the shared theta is always smaller or equal to the local theta

src/main/java/com/yahoo/sketches/theta/UpdateSketchComposition.java
Eshcar: I agree

src/main/java/com/yahoo/sketches/theta/ConcurrentUpdateSketch.java
leerho: Given any set operation on A, B: (union, intersection, difference), the "Theta rule" is 
result.theta = min( A.theta, B.theta ).  In UnionImpl line 239 and line 273, the Theta rule is applied both before and after the merge.  

What I see in your code is that you `assert getVolatileTheta() <= sketch_.getThetaLong();`, (which is not there at runtime) and at the end of the propagate you `theta_ = getThetaLong();`.   Why bother with the assert?  Just force the target theta to be the minimum of the two.  Also the `theta_ = getThetaLong();` is clearly one-sided.  It makes my head swim as to whether the Theta rule is actually being enforced.  The Theta rule IS the invariant that must be enforced, so have the code reflect that.

src/main/java/com/yahoo/sketches/theta/ConcurrentUpdateSketch.java
leerho: Yes, configurable and a default.  It is hard for us to predict the environment where this will be used.

src/main/java/com/yahoo/sketches/theta/ConcurrentThetaContext.java
leerho: What would be really useful would be a simple example (could be a test) that would not need to process much data, but serves as an example of how to set-up and use the concurrent sketch in an actual multi-threaded environment, with clear comments that explain to the user what each part of the code is doing.

src/main/java/com/yahoo/sketches/theta/ConcurrentUpdateSketch.java
leerho: I think I see another problem here, but perhaps it is because I don't completely understand what you are doing.

The Theta Union algorithm, which you appear to be emulating here, involves 3 theta values: the source, _thetaLongIn_, the destination, _gadget.getThetaLong()_, and _unionThetaLong_, which is derived from the other two.  When a new Union is created, the _unionThetaLong_ is always 1.0. 

Before the merge loop:  

*unionThetaLong_ = min(unionThetaLong_, thetaLongIn); //Theta rule*

After the merge loop:

*unionThetaLong_ = min(unionThetaLong_, gadget_.getThetaLong()); //sync thetaLongs*

Note that during the merge loop, the target gadget sketch always has integrity ... i.e., never has any dirty values .... A Theta-QuickSelect Sketch never has dirty values.  Only the Theta-Alpha sketch can have dirty values.   The Alpha sketch can never be used as a union target (gadget) only the QuickSelect sketch can.  

When the user request the result, it is always a Compact Sketch, because at this point the _unionThetaLong_ could be smaller than the gadget's thetaLong, which would invalidate (make dirty) some of the values held by the gadget.  So during the creation of the result Compact Sketch only the valid hash values are inserted into the Compact Sketch.

At no point in the processing are any of the sketches in an invalid state (with dirty values).

It is not at all clear to me how your are maintaining the integrity of the sketches during the concurrent processing.

src/main/java/com/yahoo/sketches/theta/UpdateSketchComposition.java
leerho: If we restrict the concurrent sketch to only QuickSelect theta sketches can we eliminate this?

PR #230, opened 2018-12-11
Code formatting cleanup

leerho:


src/main/java/com/yahoo/sketches/theta/ConcurrentDirectThetaSketch.java
```diff
@@ -54,39 +54,44 @@
    * Returns a (fresh) estimation of the number of unique entries
    * @return a (fresh) estimation of the number of unique entries
    */
-  @Override public double getEstimationSnapshot() {
+  @Override
```
leerho: This is probably a setting in your IDE, but the style in the library is for Override annotations to be on a separate line.  This makes the public/private modifier much easier to see as it stands out on its own line.  This should be part of Checkstyle, but isn't yet.

src/main/java/com/yahoo/sketches/theta/ConcurrentDirectThetaSketch.java
```diff
@@ -123,9 +129,9 @@
    * Ensures mutual exclusion. No other thread can update the shared sketch while propagation is
    * in progress
    */
-  @Override public void startPropagation() {
-    while (!sharedPropagationInProgress_.compareAndSet(false, true)) {
-    } //busy wait till free
+  @Override
+  public void startPropagation() {
+    while (!sharedPropagationInProgress_.compareAndSet(false, true)) { } //busy wait till free
```
leerho: Empty blocks are OK.  This formatting is simpler and easier to see that the block is, in fact, empty.

leerho:
@Eshcar 
I have already made the formatting fixes commented inline, but I wanted you to see them before this PR is merged.

I do have some more substantial design concerns.  

- Why do these concurrent classes need to be public, 
- And within these classes why do so many of the methods need to be public?  
- Ideally, only the Builder needs to be public ... and why do we need two builders?

PR #232, opened 2019-01-13
refactor to reduce visibility of concurrent classes

Eshcar:


leerho:
Why does this [constructor](https://github.com/DataSketches/sketches-core/pull/232/files#diff-13656ce699e5d1175fd06ffecd639679R56) need to be public?

When building a Direct Shared sketch, why do we need to set the bSharedIsDirect flag with a separate method?  I would think the presence of a non-null and valid WritableMemory destination would be sufficient to set it inside the method. 

The builder's public method buildLocal(UpdateSketch shared) returns a non-public ConcurrentHeapThetaBuffer.  This is a problem, we can't return to the user a non-public class. I thought we discussed returning the parent UpdateSketch.  Now, if you don't think that will work, we will have to reconsider this strategy.

I don't understand why you need methods like compactShared(), getSharedRetainedEntries(), rebuildShared(), resetShared(), sharedToByteArray(), etc, when all they do is delegate to the parent class method, which is already part of the API.

You will hate me for this, but in my last commit I reorganized the methods to group together methods that override a specific API, but no other changes other than the reordering.

Eshcar:
I changed the constructor to package visibility and buildLocal to return UpdateSketch

> When building a Direct Shared sketch, ...

I want to give the option for a HeapQS shared sketch with a (non-null) WritableMemory, as this is also the case in HQSS.

> why you need methods like compactShared() ...

ConcurrentHeapThetaBuffer works with a ConcurrentSharedThetaSketch. Any method that is not in this interface cannot be invoked by ConcurrentHeapThetaBuffer.

leerho:
UpdateSketchBuilder.setSharedIsDirect(boolean) does nothing.  Why do we need it?


> ConcurrentHeapThetaBuffer works with a ConcurrentSharedThetaSketch. Any method that is not in this interface cannot be invoked by ConcurrentHeapThetaBuffer.

You have missed my point.  Only `ConcurrentDirectThetaSketch` and `ConcurrentHeapQuickSelectSketch` implement `ConcurrentSharedThetasSketch` interface.

This interface needs to declare methods that are already existing in the target classes that implement the interface IF you need to access them from a cast to the interface.  We can use the same names, but now the only requirement is that they be public on the implementation side. Thus, all the following 12 methods in the interface can be renamed in the interface to the names in the extended class hierarchy as they are already public:

```
getSharedCurrentBytes() -> getCurrentBytes()
getSharedLowerBound() -> getLowerBound()
getSharedUpperBound() -> getUpperBound()
isSharedDirect() -> isDirect()
isSharedEmpty() -> isEmpty()
isSharedEstimationMode() -> isEstimationMode()
sharedToByteArray() -> toByteArray()
compactShared() -> compact()
compactShared(boolean, WritableMemory) -> compact(boolean, WritableMemory)
getSharedRetainedEntries() -> getRetainedEntries()
rebuildShared() -> rebuild()
resetShared() -> reset()
```
Note that none of the above methods need javadocs in the interface. (The JD output will create a link back to the extended class where there is a JD).

Only the `hashUpdate(long)` (which is not public) needs to have a different name, but I would rename it with more clarity:
`updateSingle(long) -> sharedHashUpdate(long) { hashUpdate(hash); }`

Since this method is package private, it doesn't need javadocs.

Now all 12 of the above methods can be removed from both `ConcurrentDirectThetaSketch` and `ConcurrentHeapQuickSelectSketch`.   This will remove much clutter from the two impl classes. The interface will be simplified as well without all the redundant javadocs. The "Methods for tests" comment can be removed as well.

Now that both impl classes are package private, I would rename:

`ConcurrentDirectThetaSketch` to `ConcurrentDirectQuickSelectSketch` for symmetry with the other impl class.

Eshcar:
I think I get what you are saying now.
I will work on these changes and they will be ready early next week.
Thanks.

PR #267, opened 2019-08-22
use ThreadLocalRandom instead of Random

jgeraerts:
I was using The ReservoirItemsSketch in a multithreaded program.
After replacing the normal java.util.Random by ThreadLocalRandom
I had a significant improvement in performance.

### DIFF
--- a/src/main/java/com/yahoo/sketches/sampling/ReservoirItemsSketch.java
+++ b/src/main/java/com/yahoo/sketches/sampling/ReservoirItemsSketch.java
@@ -324,8 +324,8 @@ public final class ReservoirItemsSketch<T> {
       ++itemsSeen_;
       // prob(keep_item) < k / n = reservoirSize_ / itemsSeen_
       // so multiply to get: keep if rand * itemsSeen_ < reservoirSize_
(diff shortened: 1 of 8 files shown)


yahoocla:
Thank you for submitting this pull request, however I do not see a valid CLA on file for you. Before we can merge this request please visit https://yahoocla.herokuapp.com/ and agree to the terms. Thanks! :smile:

jmalkin:
Huh, I guess we need to figure out how to disable the old Yahoo CLAbot.

The main drawback of ThreadLocalRandom is that you can't set a random seed. That's something of a mixed bag in a multithreaded environment, of course -- you don't want to have the a shared seed between threads!

And note that the data structure itself isn't thread-safe. The assumption is that you'll have independent reservoir sample sand would merge them later.

Anyway, thank you. We'll discuss this on the project's dev@ list, but I have no inherent objection.

jgeraerts:
That was the strategy we are taking. We have 1 instance of the ReservoirItemsSketch per thread but they were all reffering to the same Random instance. For us this patch was the path of least resisance.

PR #268, opened 2019-08-22
performance improvement in quantiles union

AlexanderSaydakov:
It must be faster to wrap the input here, and, as far as I can tell, should not affect the result in any way


jmalkin:
It seems like updateLogic() handles direct vs heap properly, so I agree this seems ok. I guess the previous method was the only way some (probably) single line was exercised though, hence a coverage "drop" of a might 0.006%.

PR #284, opened 2019-11-14
Minorfix format for loop.

kazuki-ma:
Let me rewrite for loop with increment/decrement.

Background
=========
Sorry to bother you but I am curious why those loops format like that.
If  you are OK let me format it.

TODO: add benchmarking.

### DIFF
--- a/src/main/java/org/apache/datasketches/HashOperations.java
+++ b/src/main/java/org/apache/datasketches/HashOperations.java
@@ -50,7 +50,7 @@ public final class HashOperations {
   public static int countPart(final long[] srcArr, final int lgArrLongs, final long thetaLong) {
     int cnt = 0;
     final int len = 1 << lgArrLongs;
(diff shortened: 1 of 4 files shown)


AlexanderSaydakov:
This change breaks the algorithm.
Why? What problem are you trying to solve?

kazuki-ma:
@AlexanderSaydakov thank you for your reply and that's my shame.

Now I'm update my PR, also update description and will add benchmark.

I'm sorry for lack of information.

leerho:
Not only is your change not equivalent, even if you corrected it, it is less efficient, depending on how smart the compiler is.   The current form "for (int i = length; i-- > 0; ) {...} " is correct and has been been around for a long long time.  You just need to learn to recognize it and understand why it is often used particularly in time sensitive loops.  

This PR is not only incorrect, it is unnecessary.

PR #294, opened 2020-01-24
Fix links: github.io to apache.org

leerho:


leerho:
For some reason the LGTM analysis on JavaScript failed !?!  There is no JavaScript!

PR #299, opened 2020-02-03
More checks for isRebuildCurMinNumKxQFlag().

leerho:


leerho:
This LGTM failure is perplexing.  There is no error log that gives any clue as to what failed.  There is no Javascript on this site.  From what I could find out by searching, apparently this can fail because there is no Javascript!!!

PR #304, opened 2020-04-03
Fix hll bounds

leerho:
This fixes the upper and lower bound estimators for HLL sketches with LgK values of 4 through 12.  The previous estimators were not completely wrong, but were off enough that justified making them much more accurate.


leerho:
Do not merge this.

leerho:
I have decided that this is still not right.  I am closing this PR.

PR #315, opened 2020-05-21
Tuple directory cleanup.

leerho:
Moved 25 ArrayOfDoubles classes from the base main/tuple package to a
new "arrayofdoubles" package.

Moved 9 ArrayOfDoubles test classes from the test/tuple package to a new
"arrayofdoubles" test package.

This required making the SerialzerDeserializer class and the Util class
in the base main/tuple package public.  Javadocs were added to those two
classes.

All the rest of the changes were required changes to import statements.

Passes all tests after this move.


jmalkin:
Looks generally ok. I dunno if either of those classes should be made non-final so they can be extended? Probably better to have new modules for specialized tuple sketches provide their own supporting classes.

PR #320, opened 2020-06-07
Renaming

leerho:
The use of "key" had multiple meanings.
In some places it meant the user supplied key and other places it meant
the hash of the user key. This was very confusing. This has been fixed.

Similarly "keys_" meant a compact array of hashes and in other places a
hash table. The arrays keys_ and summaries_ in Sketch\<S\> had overloaded
uses.  These two arrays were moved down a level to the CompactSketch and
the QuickSelectSketch to eliminate this confusion.

In the process of correcting this renaming and separation of uses, I
uncovered numerous questionable codings that could easily have caused
bugs. Certainly these confusions made the code harder to maintain and
understand.

I also added unit tests & clarified some code comments & javadocs


src/main/java/org/apache/datasketches/tuple/Sketch.java
```diff
@@ -33,10 +33,10 @@
 
   protected static final byte PREAMBLE_LONGS = 1;
 
-  long[] keys_;
-  S[] summaries_;
-  long theta_;
-  boolean isEmpty_ = true;
+  //long[] hashArr_;
+  //S[] summaryArr_;
```
davecromberge, line 37: You may have intended to remove these variables.

src/main/java/org/apache/datasketches/tuple/AnotB.java
```diff
@@ -83,31 +83,41 @@ public void notB(final Sketch<S> skB) {
...
-    final long[] hashTableKeysB = convertToHashTable(skB.keys_, countB, thetaLong_);
+    CompactSketch<S> cskB = null;
+    QuickSelectSketch<S> qskB = null;
+    final long[] hashTableB;
+    if (skB instanceof CompactSketch) {
+      cskB = (CompactSketch<S>) skB;
+      hashTableB = convertToHashTable(cskB.getHashArr(), countB, thetaLong_);
+    } else {
+      qskB = (QuickSelectSketch<S>) skB;
+      hashTableB = convertToHashTable(qskB.getHashTable(), countB, thetaLong_);
```
davecromberge, line 97: It appears that the quick select sketches' keys are already in hash table format, does this conversion merely prune out values greater than `thetaLong_`?

src/main/java/org/apache/datasketches/tuple/AnotB.java
```diff
@@ -252,44 +273,45 @@ public void notB(final org.apache.datasketches.theta.Sketch skB) {
...
     }
     //skB is not empty
     final long thetaLongB = skB.getThetaLong();
     final long thetaLong = Math.min(thetaLongA, thetaLongB);
-    //Build hashtable and removes keys of skB >= theta
     final int countB = skB.getRetainedEntries();
-    final long[] hashTableKeysB =
+
+    //Build/rebuild hashtable and removes hashes of skB >= thetaLong
+    final long[] hashTableB = //this works for all theta sketches
```
davecromberge, line 289: `//this works for all theta sketches` - it would be nice to add a little more context on what `this` is, whilst reading the documentation I couldn't figure it out.

src/main/java/org/apache/datasketches/tuple/QuickSelectSketch.java
```diff
@@ -388,96 +399,102 @@ boolean isInSamplingMode() {
...
-  S find(final long key) {
-    final int index = HashOperations.hashSearch(keys_, lgCurrentCapacity_, key);
-    if (index == -1) { return null; }
-    return summaries_[index];
-  }
+  //  S find(final long hash) {
+  //    final int index = HashOperations.hashSearch(hashTable_, lgCurrentCapacity_, hash);
+  //    if (index == -1) { return null; }
+  //    return summaryTable_[index];
+  //  }
```
davecromberge, line 425: Did you intend to exclude this?

src/main/java/org/apache/datasketches/tuple/Sketch.java
leerho: Good catch. Thanks!

src/main/java/org/apache/datasketches/tuple/AnotB.java
leerho: Yes. All Set operations must apply the "Theta Rule" whereby the resulting theta of the operation is the Min(skA, skB).  This rule is applied on line 85, where thetaLong_ is the current theta of the internal data retained by the AnotB operator for stateful operations.  Because the new thetaLong_ is potentially lower than the thetaLong of skB, we must filter out all hash values >= this new theta.  

Even though skB is in HashTable form, Open Address hash tables do not allow deletions, which means a new hash table may need to be reconstructed from the valid remaining hashes.  Whether skB is Compact or QS, we want to construct an accurate local hashTableB that we use for searching for matches.  

On closer look, I did decide that a simple optimization could be added such that if the new thetaLong_ >= thetaLongB, we can just grab a reference to the skB hashTable. Since it is only a local copy and not modified, we don't need to copy it. :)

src/main/java/org/apache/datasketches/tuple/AnotB.java
leerho: "this" refers to the following convert operation. Changed to make it clearer.  Because there are so many different flavors of Theta sketches, we need a method that works for all of them.

src/main/java/org/apache/datasketches/tuple/QuickSelectSketch.java
leerho: I did a search and it is not used anywhere.  So yes I will remove this.

leerho:
An excellent review, thank you!

src/main/java/org/apache/datasketches/tuple/AnotB.java
davecromberge: Your optimization is a smart idea and could add up over many difference operations, thanks for the explanation 👍

davecromberge:
It looks like all my comments were addressed, thanks @leerho 👍

PR #321, opened 2020-06-09
Improvements in Theta Sketch to (hopefully) eliminate the

leerho:
Misinterpretation of SingleItemSketches. 

This was driven by Hive issue Number 50.
We should try to test this in the field.

### DIFF
--- a/src/main/java/org/apache/datasketches/theta/HeapAnotB.java
+++ b/src/main/java/org/apache/datasketches/theta/HeapAnotB.java
@@ -99,10 +99,9 @@ final class HeapAnotB extends AnotB {
       Arrays.sort(compactCache);
     }
     //Create the CompactSketch
-    final boolean empty = (curCount_ == 0) && (thetaLong_ == Long.MAX_VALUE);
     final CompactSketch comp = createCompactSketch(
-        compactCache, empty, seedHash_, curCount_, thetaLong_, dstOrdered, dstMem);
-    reset();
+        compactCache, empty_, seedHash_, curCount_, thetaLong_, dstOrdered, dstMem);
+    reset(); //TODO
     return comp;
   }
 
--- a/src/main/java/org/apache/datasketches/theta/PreambleUtil.java
+++ b/src/main/java/org/apache/datasketches/theta/PreambleUtil.java
@@ -20,6 +20,8 @@
 package org.apache.datasketches.theta;
 
 import static org.apache.datasketches.Util.LS;
+import static org.apache.datasketches.Util.checkSeedHashes;
+import static org.apache.datasketches.Util.computeSeedHash;
 import static org.apache.datasketches.Util.zeroPad;
 
 import java.nio.ByteOrder;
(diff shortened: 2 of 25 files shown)


leerho:
Yes, the SingleItemSketch is never directly created by the user. It is a type of CompactSketch and results from a getResults() from a set operation, union or by a user compacting one of the updatable Theta Sketches.  Before the SIS, a normal CompactSketch with a single item consumed 24 bytes.  The SIS consumes only 16 bytes, which is a reduction of 33%.  Considering that the distribution of sketch sizes is predominantly power-law (long-tail) means if you have a millions of sketches the vast majority will have just one item.  So the reduction in storage for all of these single item sketches can be significant.

src/main/java/org/apache/datasketches/theta/Sketch.java
```diff
@@ -677,15 +677,14 @@ private static final Sketch heapifyFromMemory(final Memory srcMem, final long se
           throw new SketchesArgumentException(
               "Corrupted: COMPACT family sketch image must have Read-Only flag set");
         }
-        final boolean empty = PreambleUtil.isEmpty(srcMem); //also checks memCap < 16
-        if (empty) { //empty flag or < 16 bytes
+        final boolean empty = PreambleUtil.isEmpty(srcMem);
+        if (empty) { //TODO
```
AlexanderSaydakov, line 681: not clear what to do

src/main/java/org/apache/datasketches/theta/UnionImpl.java
```diff
@@ -431,6 +426,10 @@ private void processVer3(final Memory skMem) {
   //has seedHash and p, could have 0 entries & theta,
   // can only be compact, ordered, size >= 8
   private void processVer2(final Memory skMem) {
+    final int famId = extractFamilyID(skMem);
+    if (famId != 3) {
```
AlexanderSaydakov, line 430: should this refer to a constant in Family?

src/test/java/org/apache/datasketches/theta/EmptyTest.java
```diff
@@ -111,45 +105,6 @@ public void checkPsampling() {
     assertEquals(sk1.compact().toByteArray().length, 8);
   }
 
-  //These 3 tests reproduce a failure mode where an "old" empty sketch of 8 bytes without
```
AlexanderSaydakov, line 114: why these tests are removed?

leerho:
That's funny, I thought I replied to each of your comments, but they didn't show up.  Maybe I forgot to hit "Comment" ? :)

src/test/java/org/apache/datasketches/theta/EmptyTest.java
leerho: This was an accident.  Replaced them.

src/main/java/org/apache/datasketches/theta/UnionImpl.java
leerho: Fixed.

src/main/java/org/apache/datasketches/theta/Sketch.java
leerho: This has been fixed.

src/main/java/org/apache/datasketches/theta/PreambleUtil.java
```diff
@@ -462,6 +468,20 @@ static boolean isEmpty(final Memory mem) {
...
+  static boolean isSingleItem(final Memory mem) {
+    // Flags byte must be LittleEndian, ReadOnly, Not Empty, Compact, Ordered = 11010 = 0x1A.
+    // Flags mask will be 0x1F.
+    // SingleItem flag may not be set due to a historical bug, so we can't depend on it for now.
+    // However, if the above flags are correct, preLongs == 1, SerVer >= 3, FamilyID == 3,
+    // and the hash seed matches (not done here), it is virtually guaranteed that we have a
+    // SingleItem Sketch.
+    final boolean preLongs = extractPreLongs(mem) == 1;
+    final boolean serVer = extractSerVer(mem) >= 3;
+    final boolean famId = extractFamilyID(mem) == 3; //compact
```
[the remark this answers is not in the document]
leerho: Fixed.

src/main/java/org/apache/datasketches/theta/HeapAnotB.java
```diff
@@ -99,10 +99,9 @@ public CompactSketch getResult(final boolean dstOrdered, final WritableMemory ds
       Arrays.sort(compactCache);
     }
     //Create the CompactSketch
-    final boolean empty = (curCount_ == 0) && (thetaLong_ == Long.MAX_VALUE);
     final CompactSketch comp = createCompactSketch(
-        compactCache, empty, seedHash_, curCount_, thetaLong_, dstOrdered, dstMem);
-    reset();
+        compactCache, empty_, seedHash_, curCount_, thetaLong_, dstOrdered, dstMem);
+    reset(); //TODO
```
AlexanderSaydakov, line 104: not clear what to do here

leerho: That is a placeholder for the next round of changes.  Just ignore.
Thanks for the review!

PR #323, opened 2020-07-02
Small changes needed in master before the next set of PRs.

leerho:


src/test/java/org/apache/datasketches/tuple/aninteger/IntegerSketchTest.java
```diff
@@ -82,7 +83,7 @@ public void aNotBTest() {
       a1Sk1.update(i, 1);
     }
     anotb.update(a1Sk1, a1Sk2);
-    CompactSketch<IntegerSummary> cSk = anotb.getResult();
+    anotb.getResult();
```
AlexanderSaydakov, line 86: Why no assertions here?

leerho: I was just removing a dead assignment from the existing code.  You're right, I should of added an assertion.

PR #324, opened 2020-07-08
Refactor theta tuple

leerho:
This is a massive PR and will be difficult to review.  

- I've had to do lots of shuffling of methods between classes to place them where they make the most sense. 
- I was able to consolidate the compacting code into one new `CompactOperations` class that handles all the corner cases.  This cleaned up a lot of code that was duplicative. 
- I've extended the set operations code in both Theta and Tuple to be more flexible and with similar API. 
    - All three operators (`Union, Intersection, AnotB`) can be executed using stateful or stateless operations.  
    - This makes them easier to remember and the fact that they behave the same in both Theta and Tuple should help users. 
    - I have not yet implemented these changes in the `ArrayOfDoubles` implementation.
- Lots of documentation fixes where I came across them.  

Go ahead and review this the best you can. Independent of the reviews I want to run some characterization tests to make sure I didn't screw something up.  It passes all the unit tests plus more I added, .... but you never know :)


leerho:
I have completed the following c14n tests:
- Accuracy
- Update Speed
- Serialization Speed
- Deserialization Speed
- Estimation Speed

They all look really good.

src/main/java/org/apache/datasketches/theta/IntersectionImplR.java
```diff
@@ -82,15 +83,15 @@
...
-  }
+  //  IntersectionImplR(final short seedHash) {
+  //    seedHash_ = seedHash;
+  //    mem_ = null;
+  //    lgArrLongs_ = 0;
+  //    curCount_ = -1;
+  //    thetaLong_ = Long.MAX_VALUE;
+  //    empty_ = false;
+  //    hashTable_ = null;
+  //  }
```
davecromberge, line 94: Did you mean to remove this constructor?

src/main/java/org/apache/datasketches/BoundsOnRatiosInSampledSets.java
```diff
@@ -82,6 +82,28 @@ public static double getEstimateOfBoverA(final long a, final long b) {
...
+   */
+  public static double getEstimateOfA(final long a, final double f) {
+    checkInputs(a, 1, f);
+    return a / f;
+  }
+
+  /**
+   * Return the estimate of B. See class javadoc.
+   * @param b See class javadoc
+   * @param f the inclusion probability used to produce the set with size <i>a</i>.
```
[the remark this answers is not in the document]
leerho: You are correct! Excellent catch!

WRT Question above: The Serialization Version only relates to the storage layout and independent of the code release version.  And, yes, it our intention to retain "binary compatibility" across languages and backward compatibility across time.  We are already binary compatible with sketch binaries produced 8 years ago.   This means we can read the old sketch binaries, but when serializing, we serialize to the latest SerVer=3.

src/main/java/org/apache/datasketches/theta/AnotB.java
```diff
@@ -24,18 +24,33 @@
 
 /**
  * The API for the set difference operation <i>A and not B</i> operations.
- * This is a stateless operation. However, to make the API
- * more consistent with the other set operations the intended use is:
+ * This class include both stateful and stateless operations.
```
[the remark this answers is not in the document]
leerho: Thank you!

src/main/java/org/apache/datasketches/theta/AnotB.java
```diff
@@ -47,55 +62,125 @@ public Family getFamily() {
...
+  /**
+   * Gets the result of this operation as an ordered CompactSketch on the Java heap.
+   * @param reset If true, clears this operator to the empty state after result is returned.
+   * @return the result of this operation as a CompactSketch.
+   */
+  public abstract CompactSketch getResult(boolean reset);
 
   /**
-   * Gets the result of this set operation as a CompactSketch of the chosen form
+   * Gets the result of this stateful set operation as a CompactSketch of the chosen form. The
```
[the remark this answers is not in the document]
leerho: Fixed, thank you!

src/test/java/org/apache/datasketches/theta/UnionImplTest.java
```diff
@@ -54,14 +54,14 @@ public void checkUpdateWithSketch() {
...
-  public void checkCorruptedCompactFlag() {
+  @Test
+  public void checkUnorderedCompactFlag() {
     int k = 16;
     WritableMemory mem = WritableMemory.wrap(new byte[(k*8) + 24]);
     UpdateSketch sketch = Sketches.updateSketchBuilder().setNominalEntries(k).build();
     for (int i=0; i<k; i++) { sketch.update(i); }
     CompactSketch sketchInDirectOrd = sketch.compact(true, mem);
-    sketch.compact(false, mem); //corrupt memory
+    sketch.compact(false, mem); //change the order bit
```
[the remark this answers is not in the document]
leerho: My you have a very good eye!  At first I thought you found a bug, but what you found was an inert test, which I have now improved, thanks!  

As of version 1.3.0, there was an additional layer of classes that were ordered and unordered subclasses of the HeapCompactSketch and the DirectCompactSketch. But this was really unnecessary. So I eliminated those 4 classes as they only differed by the ordered bit.  So before, there were two ways to determine whether the CompactSketch was ordered or not: the class type, and the ordered bit (memory) or boolean (on heap).  This test was designed to test the situation where the indicators were out-of-sync and make sure an error was thrown.  

With the new version there is only one indicator of order, either a boolean on heap or a bit set in Memory. 
As a result this test was inert.  So instead of removing it, I added additional checks to make sure that the union operation would successfully and correctly merge the same hashes whether the sketch order flag was set or not.

src/test/java/org/apache/datasketches/theta/PairwiseSetOperationsTest.java
```diff
@@ -171,7 +171,7 @@ public void checkAnotBEarlyStop() {
       CompactSketch csk2 = usk2.compact(true, null);
 
       Sketch rsk = PairwiseSetOperations.aNotB(csk1, csk2);
-      double result1 = rsk.getEstimate();
+      double result1 = rsk.getEstimate(); //null ptr
```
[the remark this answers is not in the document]
leerho: These were comments I was making in the code as I was troubleshooting and forgot to go back and clean up.  The entire PairwiseSetOperations class is now deprecated since the main set operation classes now cover these pair operations as part of their API.   Eventually, when we move to version 3.0, we can delete the deprecated classes and the tests associated with them.

src/test/java/org/apache/datasketches/theta/SetOpsCornerCasesTest.java
```diff
@@ -0,0 +1,495 @@
...
+        }
+        int bytes = Sketch.getMaxCompactSketchBytes(sk.getRetainedEntries(true));
+        byte[] byteArr = new byte[bytes];
+        WritableMemory mem = WritableMemory.wrap(byteArr);
+        csk = sk.compact(false, mem);
+        break;
+      }
+    }
+    return csk;
+  }
```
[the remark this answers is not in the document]
leerho: Real good question.  I will have to look into this.

src/test/java/org/apache/datasketches/theta/HeapAlphaSketchTest.java
```diff
@@ -291,9 +290,9 @@ public void checkAlphaToCompactForms() {
     assertEquals(comp4.getUpperBound(2), comp1ub);
     assertEquals(comp4.isEmpty(), false);
     assertEquals(comp4.isEstimationMode(), estimating);
-    assertEquals(comp1bytes, comp4.getCurrentBytes(true)); //flag is not relevant
+    assertEquals(comp1bytes, comp4.getCompactBytes()); //flag is not relevant
```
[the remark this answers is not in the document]
leerho: thanks!

src/test/java/org/apache/datasketches/theta/CompactSketchTest.java
```diff
@@ -222,22 +216,80 @@ public void checkMemTooSmallOrdered() {
...
+    csk2o = csku.compact(true, null); //single, heap, ordered
+    state.check(csk2o);
+
+    csk2o = csku.compact(false, null); //single, heap, ordered
+    state.check(csk2o);
+
+    csk2o = csko.compact(true, null); //single, heap, ordered
+    state.check(csk2o);
+
+    csk2o = csko.compact(false, null); //single, heap, ordered
```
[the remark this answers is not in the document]
leerho: Thanks!

src/main/java/org/apache/datasketches/theta/IntersectionImplR.java
leerho: Yes.  It was placed there at one time when we thought we needed it.  But nothing calls it, it is dead code.  I left it there in case I thought of another reason to put it back.

src/main/java/org/apache/datasketches/tuple/CompactSketch.java
```diff
@@ -121,6 +121,11 @@
     }
   }
 
+  @Override
+  public CompactSketch<S> compact() {
+    return this;
+  }
```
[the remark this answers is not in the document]
leerho: In order to do that we would have to add a method getSerializationBytes() or equivalent to the Summary Interface.  This is not a bad idea. Let me think about it.

src/main/java/org/apache/datasketches/BoundsOnRatiosInSampledSets.java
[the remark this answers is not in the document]
davecromberge: Thank you for addressing my concerns about serialisation - your answer also cleared my misconceptions about the library version being tied to the serialisation.

src/test/java/org/apache/datasketches/theta/SetOpsCornerCasesTest.java
```diff
@@ -0,0 +1,495 @@
...
+  @Test
+  //Check all corner cases against standard Union, Intersection, and AnotB.
+  //The unordered case is not tested
+  public void compareCornerCases() {
+    int k = 64;
+    for (State stateA : State.values()) {
+      for (State stateB : State.values()) {
+        if ((stateA == EST_MEMORY_UNORDERED) || (stateB == EST_MEMORY_UNORDERED)) { continue; }
+        cornerCaseChecks(stateA, stateB, k);
+        //cornerCaseChecksMemory(stateA, stateB, k);
```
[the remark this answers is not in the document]
leerho: No I did not.  Thanks!

src/test/java/org/apache/datasketches/theta/SetOpsCornerCasesTest.java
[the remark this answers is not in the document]
leerho: Actually we do.  But the corner-case tests are executed via one of the concrete implementations and in this case in the `../tuple/adouble` package.  And because of the way generics work, we only need to exercise these tests from one of the concrete implementations.  The way I wrote them in the `adouble`package is probably a better way to do it, and I haven't rewritten the tests in theta yet to take advantage of this.

src/test/java/org/apache/datasketches/theta/UnionImplTest.java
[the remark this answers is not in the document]
davecromberge: It makes sense to remove the various implementations of the different variants.  When performing set operations between sets with different order bits, are the unordered sets sorted first internally?  I ask this because searching the internal hash keys should still work in theory, if the keys are unsorted.

src/test/java/org/apache/datasketches/theta/SetOpsCornerCasesTest.java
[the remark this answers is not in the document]
davecromberge: Thanks for looking into this - I will take a look at the tests in the `adouble` package.  Having the same tests for Tuple is re-assuring, since I am using more and more Tuple set operations in my use cases.

src/main/java/org/apache/datasketches/tuple/CompactSketch.java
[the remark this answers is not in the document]
davecromberge: Great - just to add some detail - it is possible now to compute this by serialising the sketch and measuring the resulting byte size.  However, this is obviously wasteful and my question above is only predicated on whether the internal bytes are already available.

src/test/java/org/apache/datasketches/theta/UnionImplTest.java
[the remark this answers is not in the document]
leerho: In an UpdateSketch the hashes are in a Knuth Open Address Double Hash (OADH) power-of-2 sized array; essentially a hash table, but a very efficient one.  The placement of hashes in a hash table is random, so the hashes are never in any order. The CompactSketch has all the same hashes, but all the empty slots of the hash table have been removed.  When creating the CS, the user has the option to have the hashes sorted or not. Sorting costs a little more time when creating the CS, but when merging CSs into a union, we can take advantage of "early-stop" which results in dramatic improvement in merge speed performance.  We never "search" the hashes in a CS. There is no need to do that.

The CS is the simplest representation of a sketch.  It contains primarily the hashes and Theta, that's it!  It is immutable and has no concept of "K" or "Nominal Entries".  Yet it can be used as input of all set operations.

src/main/java/org/apache/datasketches/tuple/CompactSketch.java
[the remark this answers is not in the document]
leerho: Clearly doing toByteArray() would be very wasteful. The only knowledge we have of the Summary is the two current methods: we can copy it and serialize it to bytes.  When we are serializing the sketch we have to call toByteArray() anyway, so we grab its size at that point.  

But adding this method would give us that visibility and I think it could be useful.  However, we don't want to make the requirements on the Summary too restrictive as it may limit what the users want to do with it.  

Perhaps the best way to resolve this would be to put a [DISCUSS] issue on this topic on our dev@ list and see if anyone responds pro or con.

leerho:
An excellent review, thanks to David.  Clearly this PR is not done.  I will be submitting corrections to resolve these issues soon.

src/main/java/org/apache/datasketches/tuple/CompactSketch.java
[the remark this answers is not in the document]
leerho: This is the only item that is not resolved yet.

leerho:
This failure is really odd.  I cannot reproduce it locally either in Eclipse or from Maven.

davecromberge:
Lee, I've just done a fresh checkout of this branch on my machine, and run this command in my terminal:

`mvn clean compile test -Dgpg.skip=true`

I get the same failure as the build server.  I am using Adopt OpenJDK version 8.0.252.j9-adpt.  I'll try and open in my IDE and see if I can also reproduce it there.

leerho:
The 3 bugs:
- The EMPTY_SKETCH_MASK in EmptyCompactSketch was not masking out the lgNomLongs and lgArrLongs fields. This would cause it to reject an otherwise valid empty compact sketch.
- The `loadCompactMemory()` as part of `CompactOperations` (this is new) was not initializing the `lgNomLongs` and `lgArrLongs` fields to zero. This contributed to the 1st bug, especially when using off-heap memory.
- The new centralized `memoryToCompact()` as part of `CompactOperations` was not properly loading the flags field into the destination compact sketch.

src/main/java/org/apache/datasketches/theta/AnotB.java
```diff
@@ -47,55 +62,174 @@ public Family getFamily() {
...
+   * a multistep operation with multiple <i>notB</i> steps. Other following steps can still produce
+   * a valid result.</p>
+   *
+   * <p>Use {@link #getResult(boolean)} to obtain the result.</p>
    *
-   * @return the result of this set operation as a CompactSketch of the chosen form
+   * @param skB The incoming Tuple sketch for the second (or following) argument <i>B</i>.
    */
-  public abstract CompactSketch getResult(boolean dstOrdered, WritableMemory dstMem);
+  public abstract void notB(Sketch skB);
```
davecromberge, line 108: This additional documentation is great, especially the clarity provided for exception cases vs legal empty sets.  There are two references to a Tuple sketch, perhaps you mean `Theta` instead?

src/main/java/org/apache/datasketches/theta/CompactOperations.java
```diff
@@ -229,10 +233,20 @@ static final Memory loadCompactMemory(
...
+    // As written below, all fields are initialized avoiding an extra copy.
+
+    //The first 8 bytes (pre0)
     insertPreLongs(dstMem, preLongs); //RF not used = 0
     insertSerVer(dstMem, SER_VER);
     insertFamilyID(dstMem, famID);
-    //ignore lgNomLongs, lgArrLongs bytes for compact sketches
+    //The following initializes the lgNomLongs and lgArrLongs to 0.
+    //They are not used in CompactSketches.
+    dstMem.putShort(LG_NOM_LONGS_BYTE, (short)0);
```
davecromberge, line 249: Does clearing LG_NOM_LONGS_BYTE to zero have the effect of clearing both lgNomLongs and lgArrLongs?  From the preamble output, it appeared that they were two separate fields, and I didn't see an explicit byte cleared for lgArrLongs.

src/main/java/org/apache/datasketches/theta/HeapAlphaSketch.java
```diff
@@ -128,20 +128,19 @@ static HeapAlphaSketch heapifyInstance(final Memory srcMem, final long seed) {
...
     final double alpha = nomLongs / (nomLongs + 1.0);
     final long split1 = (long) (((p * (alpha + 1.0)) / 2.0) * LONG_MAX_VALUE_AS_DOUBLE);
 
-    if ((myRF == ResizeFactor.X1)
-            && (lgArrLongs != startingSubMultiple(lgNomLongs + 1, myRF, MIN_LG_ARR_LONGS))) {
-      throw new SketchesArgumentException("Possible corruption: ResizeFactor X1, but provided "
-              + "array too small for sketch size");
+    if (isResizeFactorIncorrect(srcMem, lgNomLongs, lgArrLongs)) {
+      throw new SketchesArgumentException("Possible corruption: ResizeFactor  "
+          + "inconsistent with lgNomLongs and lgArrLongs.");
```
davecromberge, line 140: This is a different approach taken than that of the DirectQuickSelect sketch, where X2 is selected as a resize factor when the arguments are incorrect, as per the comment:
```//If incorrect it sets it to X2 which always works.```
Is this by design?

File: src/test/java/org/apache/datasketches/theta/CompactSketchTest.java [hunk 15]
```diff
@@ -333,6 +333,55 @@ public void checkHeapCompactSketchCompact() {
...
+
+    //exact
+    sk.update(2);
+    bytes = sk.getCompactBytes();
+    wmem1 = WritableMemory.allocate(bytes);
+    wmem2 = WritableMemory.allocate(bytes);
+    csk1 = sk.compact(true, wmem1); //place into memory
+    assertTrue(csk1 instanceof DirectCompactSketch);
+    csk2 = csk1.compact(true, wmem2);
+    assertTrue(csk2 instanceof DirectCompactSketch);
```
davecromberge, line 371: What is the significance of the additional value, when compared to a sketch containing a single value?  Considering that lgK = 6, it is clearly still in exact mode, is the intention here to see that the additional value serialises properly into memory?

src/main/java/org/apache/datasketches/tuple/CompactSketch.java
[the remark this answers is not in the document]
davecromberge: I have looked through the additional changes, and yes, all the rest of the points raised were resolved.  Thank you for addressing them.  I have also read through the code related to the bug fixes mentioned and the additional commentary on clearing memory is valuable.

Concerning discussion thread here, I am interested in how the summary may be restricted.  Were you considering adding an additional method to the summary type to provide the size for the tuple?  
If it is more involved, I am happy to pick this up on [DISCUSS] or not let this block the PR.

src/main/java/org/apache/datasketches/theta/CompactOperations.java
leerho: lgNomLongs is a byte followed by lgArrLongs, which is also a byte.  2 bytes = 1 short.  so clearing the address of lgNomLongs clears both.  The power of Memory (and its danger, if you don't know what you are doing) is that it acts like a combination of C struct and C union so that you can write 8 bytes as a byte[] and then read it back as a long or a double.

src/main/java/org/apache/datasketches/theta/HeapAlphaSketch.java
leerho: Yes.  Look more closely at the object being modified in the case where X2 is set.  It is a WritableMemory.  In the case where  I have to throw, it is just Memory, which is immutable.  Big difference.

src/test/java/org/apache/datasketches/theta/CompactSketchTest.java
leerho: Yes. But I need to add more asserts to make the difference between these cases more clear.

src/main/java/org/apache/datasketches/theta/AnotB.java
leerho: Good catch.  This is a copy/paste error. Thank you!

src/main/java/org/apache/datasketches/theta/AnotB.java
```diff
@@ -47,55 +62,174 @@ public Family getFamily() {
...
-   * @return the result of this operation as an ordered CompactSketch on the Java heap
+   * This is part of a multistep, stateful AnotB operation and sets the given Theta sketch as the
+   * first argument <i>A</i> of <i>A-AND-NOT-B</i>. This overwrites the internal state of this
+   * AnotB operator with the contents of the given sketch.
+   * This sets the stage for multiple following <i>notB</i> steps.
+   *
+   * <p>An input argument of null will throw an exception.</p>
+   *
+   * <p>Rationale: In mathematics a "null set" is a set with no members, which we call an empty set.
+   * That is distinctly different from the java <i>null</i>, which represents a nonexistant object.
```
AlexanderSaydakov, line 73: nonexistent

src/main/java/org/apache/datasketches/theta/AnotB.java
```diff
@@ -47,55 +62,174 @@ public Family getFamily() {
...
+   *
+   * <p>This a stateless operation and has no impact on the internal state of this operator.
+   * Thus, this is not an accumulating update and does not interact with the {@link #setA(Sketch)},
+   * {@link #notB(Sketch)}, {@link #getResult(boolean)}, or
+   * {@link #getResult(boolean, WritableMemory, boolean)} methods.</p>
+   *
+   * <p>If either argument is null an exception is thrown.</p>
+   *
+   * <p>Rationale: In mathematics a "null set" is a set with no members, which we call an empty set.
+   * That is distinctly different from the java <i>null</i>, which represents a nonexistant object.
```
AlexanderSaydakov, line 178: nonexistent

src/main/java/org/apache/datasketches/theta/AnotBimpl.java
```diff
@@ -0,0 +1,266 @@
...
+    CompactSketch cskB = null;
+    UpdateSketch uskB = null;
+    final long[] hashTableB;
+    if (skB instanceof CompactSketch) {
+      cskB = (CompactSketch) skB;
+      hashTableB = convertToHashTable(cskB.getCache(), countB, thetaLong_, REBUILD_THRESHOLD);
+    } else {
+      uskB = (UpdateSketch) skB;
+      hashTableB = (thetaLong_ < thetaLongB)
+          ? convertToHashTable(uskB.getCache(), countB, thetaLong_, REBUILD_THRESHOLD)
```
AlexanderSaydakov, line 105: why convert hash table to another hash table? theta check can be done before looking up in existing one

src/main/java/org/apache/datasketches/theta/AnotBimpl.java
```diff
@@ -0,0 +1,266 @@
...
+  public CompactSketch getResult(final boolean reset) {
+    return getResult(true, null, reset);
+  }
+
+  @Override
+  public CompactSketch getResult(final boolean dstOrdered, final WritableMemory dstMem,
+      final boolean reset) {
+    final CompactSketch result =
+        CompactOperations.componentsToCompact(
+            thetaLong_, curCount_, seedHash_, empty_, true, false, dstOrdered, dstMem, hashArr_);
```
AlexanderSaydakov, line 141: it seems that hashArr_ is not a copy, which it should be unless reset is requested

src/main/java/org/apache/datasketches/theta/SetOperationBuilder.java
```diff
@@ -207,11 +207,11 @@ public SetOperation build(final Family family, final WritableMemory dstMem) {
       }
       case A_NOT_B: {
         if (dstMem == null) {
-          setOp = new HeapAnotB(bSeed);
+          setOp = new AnotBimpl(bSeed);
         }
         else {
-          throw new SketchesArgumentException(
-            "AnotB is a stateless operation and cannot be persisted.");
+          throw new SketchesArgumentException( //TODO we should be able to do this now.
```
AlexanderSaydakov, line 213: is this TODO comment still relevant?

src/main/java/org/apache/datasketches/theta/AnotB.java
leerho: Thank you. Actually found this spelling error in 6 places :)

src/main/java/org/apache/datasketches/theta/CompactOperations.java
davecromberge: I missed that on first inspection - thanks for explaining.  In the context of a performance critical piece of code, having this low level implementation makes sense.

src/main/java/org/apache/datasketches/theta/AnotBimpl.java
leerho: Good catch. Thank you! Will fix.

src/main/java/org/apache/datasketches/theta/AnotBimpl.java
leerho: Good catch.  Fixed.

src/main/java/org/apache/datasketches/theta/SetOperationBuilder.java
leerho: Removed the TODO, but still not allowing AnotB to be persisted.

PR #329, opened 2020-08-04
Rename "simpleIntLog2" to "simpleLog2OfLong" plus javadoc changes.

leerho:


src/main/java/org/apache/datasketches/Util.java
```diff
@@ -452,14 +453,36 @@ public static int ceilingPowerOf2(final int n) {
...
+   * <li>n == a power of 2 : returns n</li>
+   * <li>otherwise returns the largest power of 2 less than n and equal to a mathematical
+   * integer.</li>
+   * </ul>
+   *
+   * @param n The given long argument.
+   * @return the floor power of 2 as a long
+   */
+  public static long floorPowerOf2(final long n) {
+    if (n <= 1) { return 1; }
```
davecromberge, line 482: Could returning a sentinel value of 1 lead to unexpected errors for n less than or equal to one?  I am curious why the zeroth power is not returned instead for this case - is 1 used for identity purposes?

src/main/java/org/apache/datasketches/Util.java
```diff
@@ -452,14 +453,36 @@ public static int ceilingPowerOf2(final int n) {
...
    */
   public static int floorPowerOf2(final int n) {
     if (n <= 1) { return 1; }
     return Integer.highestOneBit(n);
   }
 
+  /**
+   * Computes the floor power of 2 given <i>n</i> is in therange [1, 2^63-1].
+   * This is the largest positive power of 2 that equal to or less than the given n and equal
+   * to a mathematical integer.
```
davecromberge, line 467: There is a slight typo in the javadoc for this and the previous method:
```This is the largest positive power of 2 that _is_ equal```

File: src/main/java/org/apache/datasketches/Util.java (hunk 1 above)
Comment (leerho) [reply src/main/java/org/apache/datasketches/Util.java:482]: This function returns a value that is the largest integer power of 2 <= n for the range [1, 2^63-1].  So if I enumerate:

```
n  result
1  1  //2^0
2  2  //2^1
3  2
4  4  //2^2
5  4
6  4
7  4
8  8  //2^3
etc.
```

0 is not power of 2, nor is any negative value.    
There are powers of two that are fractions < 1.0:  

```
.5   = 2^(-1)
.25 = 2^(-2)
etc.,
```

but these cannot be expressed as an integer.

src/main/java/org/apache/datasketches/Util.java
leerho: Thanks fixed.

src/main/java/org/apache/datasketches/Util.java
leerho: I think you may be confusing a number that is a power of 2 vs a number that is an exponent of 2.

davecromberge: Yes, my mistake, I did confuse the two. Thanks for clarifying!

PR #348, opened 2021-02-08
Made parallel APIs for SetOperations in Theta, and Tuple (generics).

leerho:
Made parallel APIs for SetOperations in Theta, and Tuple (generics): We now have both stateful and stateless operations for union, intersection and AnotB.  Added comprehensive examples (and live tests) for these added stateless and stateful operations.

Made consistent naming of union and intersection operations in ArrayOfDoubles package. We don't have stateless operations for union and intersection in ArrayOfDoubles. This can wait for when we do major refactoring of that package.

Also added more complete testing for the inequality search methods for long arrays, which were missing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@
 
 =================
 
-# Apache&reg; DataSketches&trade; Core Java Library Component
+# Apache<sup>&reg;</sup> DataSketches&trade; Core Java Library Component
 This is the core Java component of the DataSketches library.  It contains all of the sketching algorithms and can be accessed directly from user applications. 
 
 This component is also a dependency of other components of the library that create adaptors for target systems, such as Hadoop Pig and Hadoop Hive.
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@
 
 =================
 
-# DataSketches Core Java Library Component
+# Apache&reg; DataSketches&trade; Core Java Library Component
 This is the core Java component of the DataSketches library.  It contains all of the sketching algorithms and can be accessed directly from user applications. 
 
 This component is also a dependency of other components of the library that create adaptors for target systems, such as Hadoop Pig and Hadoop Hive.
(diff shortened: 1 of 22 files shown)


davecromberge (commented):
These changes look good 👍 

I have a question about the use of `reset` to reset the state within the stateless intersection and union operations.  Specifically, was this related to the bug that you found in the release candidate?  
How likely is it that same union/intersection instance could be used in some concurrent context, like an RDD in Spark?  This question seems to have come up in the mailing list - is it worth putting one or two lines of java doc in the SetOperation base class?  Otherwise, resets may interleave.

davecromberge (approved):

leerho:
"Stateless" does not imply concurrent.  It just means everything is completed in one action.  The term "stateless" is also used to contrast it with "stateful" where the sketch object holds state between subsequent calls to the same method. The user must be aware of this stateful property and respect it.   "Stateless" also does not imply that "stateless" and "stateful" operations can be interleaved.  

All of our sketches are non-thread-safe, with the exception of the "Concurrent*ThetaSketch" family, which was specifically designed to be thread-safe.

Perhaps this comment should be in the documentation somewhere?

leerho:
We already have [a clear statement of "not-thread-safe"](https://datasketches.apache.org/docs/Architecture/Concurrency.html) in our documentation ... do we need to say more?

davecromberge:
> We already have a clear statement of "not-thread-safe" in our documentation ... do we need to say more?

I agree with you - you raise a good point that "stateless" does not imply concurrent.  This is just something that I thought of in passing, and admittedly it's a contrived scenario, based on one or two that may have come up on the mailing list.  Personally, I don't see any reason to reuse a set operation builder, and in our use case we construct a virgin instance for every union or intersection operation, despite serving many concurrent requests.

The documentation you have linked to is clear and explicit about the concerns raised in my comment.  Thanks for. your response and additional detail.

AlexanderSaydakov (approved):

PR #352, opened 2021-04-19
Fix warnings.

leerho:
Very minor fixes to clean up warnings.


jmalkin (approved):
👍 

verified that intellij knows that "restrictions" warning

PR #363, opened 2021-09-18
Update datasketches-memory 2.0.0

davecromberge:
Uses new datasketches memory 2.0.0 which has added support for Java9 - 13.


leerho:
"Coveralls API failed with HTTP status 405: Not Allowed (Response doesn't contain Content-Type header)"

Hmm. What is this about?  I haven't seen this one before.

leerho:
Retried running the tests. Still fails.

leerho:
Coveralls is down for maintenance.  Perhaps that is the problem.

davecromberge:
In addition to the coveralls failure - I've noticed that the matrix tests have been failing intermittently lately both here and in datasketches-memory.  The issue seems to be related to an aggressive timeout setting on downloading the super pom.  I'd like to investigate extending the timeout setting in the GH actions to make the tests more reliable.

leerho:
Coveralls site is back up, but this job still fails with error 405 (Not allowed).

davecromberge:
@leerho the checks have now passed successfully - I think that the issues were related to the coveralls maintenance.  

As for the other transient issue that I mentioned with the super pom, I will have to keep an eye out for any failures and potentially increase HTTP timeouts via maven arguments.  My preference is to monitor the situation instead of adding complexity to the GH actions build.

leerho:
Agreed.

PR #365, opened 2021-09-21
Test coveralls on master

davecromberge:


davecromberge:
This is to eliminate Memory2 as a potential cause for coveralls reporting problems.

PR #367, opened 2021-09-28
Remove deprecated methods

davecromberge:
Release v3.0.0 preparation

Since this is a major release, methods marked with a deprecation notice are now removed.


davecromberge:
Unfortunately this task is not complete.  I was unable to definitively remove PairwiseSetOperations because there was some corner case testing that looks important.  Also, there are some tests that have "Deprecated" in the case name but I cannot determine why.

leerho (approved):

PR #374, opened 2021-11-18
Simplified set operations

AlexanderSaydakov:
removed NewDegen state, renamed New to Empty, some rearrangement


leerho (approved):

leerho:
Thank you for doing this.

PR #376, opened 2021-12-03
Fix set op leakage

leerho:
This fixes a few places where I found the potential for parameter leakage.


leerho:
These are pretty straight forward fixes.  I'm going to go ahead and push.

PR #377, opened 2021-12-03
Updated SetOperationCornerCases for Unions and cleaned up the code.

leerho:


src/main/java/org/apache/datasketches/tuple/Union.java
```diff
@@ -134,64 +130,78 @@ public void union(final org.apache.datasketches.theta.Sketch thetaSketch, final
...
-    if (thetaLong_ >= qsk_.thetaLong_ && qsk_.getRetainedEntries() <= qsk_.getNominalEntries()) {
-      return qsk_.compact();
-    }
-    long theta = min(thetaLong_, qsk_.thetaLong_);
+      result = qsk_.compact();
+    } else if (thetaLong_ >= qsk_.thetaLong_ && qsk_.getRetainedEntries() <= qsk_.getNominalEntries()) {
+      result = qsk_.compact();
+    } else {
+      long theta = min(thetaLong_, qsk_.thetaLong_);
+      int numHashes = 0;
```
davecromberge, line 167: Could you please explain the second "if" branch with regard to the state combination truth table?  From some of the conditional checks, it is clear that `qsk` is an exact mode sketch.  It is also clear from the first branch that the local state is not empty.  Where the difficulty arises is that the state for `qsk` and `thetaLong` is non-local to this function and requires more context.

davecromberge (approved):

src/main/java/org/apache/datasketches/tuple/Union.java
leerho: This is confusingly written, I admit. I am looking into how it can be made more direct.  While studying this, I may have found another problem, which I need to investigate.

leerho: I have cleaned up this section and added comments so that I could figure out what exactly was going on :)

src/test/java/org/apache/datasketches/theta/CornerCaseThetaSetOperationsTest.java
```diff
@@ -20,6 +20,8 @@
 package org.apache.datasketches.theta;
 
 import org.testng.annotations.Test;
+//import static org.apache.datasketches.Util.DEFAULT_UPDATE_SEED;
+//import static org.apache.datasketches.hash.MurmurHash3.hash;
```
AlexanderSaydakov, line 25: do we need to keep these commented-out lines?

src/main/java/org/apache/datasketches/tuple/Union.java
davecromberge: The additional comments were helpful and it makes more sense now - thanks!

leerho:
I'm going to add to this the corner-case tests for the ArrayOfDoubles Tuple Sketch.

src/test/java/org/apache/datasketches/theta/CornerCaseThetaSetOperationsTest.java
leerho: These two imports are associated with some helper methods at the bottom of the file, which are also commented out.  These helper methods would aid a reader in understanding how the various test hash values used in all the tests are actually generated.

leerho:
Lots of changes in this PR.  A lot of changes in var names to make the code more understandable.  Also lots of changes in javadocs and code comments.  But the code runs all the regular tests plus all of the corner case tests for theta, tuple (all the generics) and the ArrayOfDoubles Tuple sketches.  Code coverage is still above 96%. So unless something comes up in the review, it should be ready for a release.  

I did discover a problem with Backward compatibility with SerVer 1 theta sketches that works (in a strange way) internally in Yahoo, but was not done correctly.  Once Will has moved his code to V3, we can deprecate all the compatibility with V1 and V2.  These old Serialization Versions have never appeared in Open Source code.  We should be able to make this change sometime in Q1/2022. 

Please Review, 

Thanks

AlexanderSaydakov (approved):

PR #378, opened 2021-12-22
Code additions requested by Will.

leerho:
Added 4 methods to Sketches class:

heapifyCompactSketch(Memory)
heapifyCompactSketch(Memory, long)
wrapCompactSketch(Memory)
wrapCompactSketch(Memory, long)

Changed the behavior of heapifySketch(2) and wrapSketch(2) calls
so that if the given image is a CompactSketch the behavior will
be the same as above.

The basic behavior change is that a seed is no longer required to
heapify or wrap a CompactSketch image.
But if it is given, it will be used to check the
hashSeed of the image, if there is one.

This push represents the main code changes.
More test code needs to be added.

### DIFF
--- a/.mvn/jvm.config
+++ b/.mvn/jvm.config
@@ -1,1 +1,0 @@
---add-opens java.base/jdk.internal.ref=ALL-UNNAMED
--- a/pom.xml
+++ b/pom.xml
@@ -698,6 +698,7 @@ under the License.
                 <argLine>
                   --add-exports java.base/jdk.internal.misc=ALL-UNNAMED
                   --add-exports java.base/jdk.internal.ref=ALL-UNNAMED
+                  --add-opens java.base/jdk.internal.ref=ALL-UNNAMED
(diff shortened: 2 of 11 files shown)


leerho:
If you want to run this branch from the command line, I recommend:
`mvn clean test -Djdk.version=1.8`
It will run in JDK 9+, but it needs a JVM arg: 
`--add-opens java.base/jdk.internal.ref=ALL-UNNAMED`
I still figuring out the best way to specify that and whether to use
- .mvn/jvm.config
- MAVEN_OPTS
- settings.xml
- or via Command line: mvn clean test -Dargline=""
- etc.
-

leerho:
If you are running JDK9+ with an IDE, you will need to add the same "--add-opens ..." to either IntelliJ or Eclipse.

leerho:
Could not get LGTM to build the base directory. I'm going to have to close this and reopen a new PR.

PR #379, opened 2021-12-23
Add heapify wrap compact sketch2

leerho:


leerho:
Added 4 methods to Sketches class:

heapifyCompactSketch(Memory)
heapifyCompactSketch(Memory, long)
wrapCompactSketch(Memory)
wrapCompactSketch(Memory, long)

Changed the behavior of heapifySketch(2) and wrapSketch(2) calls
so that if the given image is a CompactSketch the behavior will
be the same as above.

The basic behavior change is that a seed is no longer required to
heapify or wrap a CompactSketch image.
But if it is given, it will be used to check the
hashSeed of the image, if there is one.

This push represents the main code changes.
More test code needs to be added.

If you want to run this branch from the command line, I recommend:
mvn clean test -Djdk.version=1.8
It will run in JDK 9+, but it needs a JVM arg:
--add-opens java.base/jdk.internal.ref=ALL-UNNAMED
I'm still figuring out the best way to specify that and whether to use

- .mvn/jvm.config
- MAVEN_OPTS
- settings.xml
- or via Command line: mvn clean test -Dargline=""
- etc.

If you are running JDK9+ with an IDE, you will need to add the same "--add-opens ..." to either IntelliJ or Eclipse

leerho:
We have a test that is failing randomly, probably due a random generator not being made deterministic:
This will need to be fixed before we can release.  It has nothing to do with this PR.

src/test/java/org.apache.datasketches.sampling/
ReservoirLongsSketchTest.checkEstimateSubsetSum:461 expected [true] but found [false].

leerho:
Note this should compile & test from the Maven command line, but there is currently a bug in eclipse that forces you to use a workaround if you try to compile in Java 9+.

src/test/java/org/apache/datasketches/theta/SketchTest2.java
```diff
@@ -0,0 +1,286 @@
...
+package org.apache.datasketches.theta;
+
+import static org.apache.datasketches.Util.DEFAULT_UPDATE_SEED;
+import static org.testng.Assert.assertEquals;
+
+import org.apache.datasketches.memory.Memory;
+import org.apache.datasketches.tuple.Util;
+import org.testng.annotations.Test;
+
+public class SketchTest2 {
```
AlexanderSaydakov, line 29: Could we have a more meaningful name please?

leerho: OK, I renamed the test class as you requested. However, in looking at the tests I realized that I had all the heapify tests, but I forgot to add the "wrap" tests.  So I added them.

AlexanderSaydakov (approved):

src/main/java/org/apache/datasketches/theta/SingleItemSketch.java
```diff
@@ -99,6 +101,21 @@ public static SingleItemSketch heapify(final Memory srcMem, final long seed) {
...
   }
 
+  /**
+   * Creates a SingleItemSketch on the heap given a SingleItemSketch Memory image and a seedHash.
+   * Checks the seed hash of the given Memory against a hash of the given seed.
+   * @param srcMem the Memory to be heapified.
+   * @param seedHash a given seedHash
+   * @return a SingleItemSketch
+   */ //does not override Sketch
+  public static SingleItemSketch heapify(final Memory srcMem, final short seedHash) {
```
will-lauer, line 111: it would probably be clearer here if `seedHash` was renamed `expectedSeedHash` as the actual seed has is extracted from the `srcMem` object.

src/main/java/org/apache/datasketches/theta/CompactSketch.java
```diff
@@ -19,25 +19,241 @@
...
+   *
+   * <p>This this method checks if the given seed was used to create the source Memory image.
+   * However, SerialVersion 1 sketches cannot be checked.</p>
+   *
+   * @param srcMem an image of a CompactSketch that was created using the given seed.
+   * <a href="{@docRoot}/resources/dictionary.html#mem">See Memory</a>.
+   * @param seed <a href="{@docRoot}/resources/dictionary.html#seed">See Update Hash Seed</a>.
+   * @return a CompactSketch on the heap.
+   */
+  public static CompactSketch heapify(final Memory srcMem, final long seed) {
```
will-lauer, line 104: we should think about renaming `seed` here to `expectedSeed`, as mostly seed is being used here as a validation of the seedHash that is in the `srcMem`, and not setting the seed to be a specific value (v1 sketches excepted). Even though I know what this was supposed to do, I got confused reading the code because I was thinking this was passing in the seed  rather than specifying the expected value to validate against.

src/main/java/org/apache/datasketches/theta/DirectCompactSketch.java
```diff
@@ -60,16 +60,16 @@
...
    * @param srcMem <a href="{@docRoot}/resources/dictionary.html#mem">See Memory</a>
-   * @param seed The update seed.
-   * <a href="{@docRoot}/resources/dictionary.html#seed">See Update Hash Seed</a>.
+   * @param seedHash The update seedHash.
+   * <a href="{@docRoot}/resources/dictionary.html#seedHash">See Seed Hash</a>.
    * @return this sketch
    */
-  static DirectCompactSketch wrapInstance(final Memory srcMem, final long seed) {
-    checkMemorySeedHash(srcMem, seed);
+  static DirectCompactSketch wrapInstance(final Memory srcMem, final short seedHash) {
```
will-lauer, line 67: `seedHash` here should probably be renamed to `expectedSeedHash` to clarify its use.

src/main/java/org/apache/datasketches/theta/Sketch.java
```diff
@@ -57,125 +53,140 @@
...
+   * <a href="{@docRoot}/resources/dictionary.html#mem">See Memory</a>.
    * @return a Sketch backed by the given Memory
    */
   public static Sketch wrap(final Memory srcMem) {
-    return wrap(srcMem, DEFAULT_UPDATE_SEED);
+    final int  preLongs = srcMem.getByte(PREAMBLE_LONGS_BYTE) & 0X3F;
+    final int serVer = srcMem.getByte(SER_VER_BYTE) & 0XFF;
+    final int familyID = srcMem.getByte(FAMILY_BYTE) & 0XFF;
+    final Family family = Family.idToFamily(familyID);
+    if (family == Family.QUICKSELECT) {
```
will-lauer, line 134: Is it possible to have family == ALPHA here?

will-lauer (approved):
I think it may be worth changing the names of some parameters so that its obvious where seeds are being used to validate rather than set the seed in use, but otherwise I don't seen any problems.

src/main/java/org/apache/datasketches/theta/SingleItemSketch.java
leerho: This is a good suggestion.  So I made this variable name change everywhere applicable (see comments on latest push).

src/main/java/org/apache/datasketches/theta/CompactSketch.java
leerho: Same comment as above.

src/main/java/org/apache/datasketches/theta/DirectCompactSketch.java
leerho: Same comment as above.

src/main/java/org/apache/datasketches/theta/Sketch.java
leerho: No. The Alpha sketch was never intended to be wrapped.

PR #380, opened 2022-01-07
In executing Clover in prep for release I found some obsolete classes in

leerho:
AoD that were totally not used (probably my bad!).

I also found some critical missing functionality at the root Sketch
level. So I added that.


AlexanderSaydakov (approved):

src/main/java/org/apache/datasketches/tuple/arrayofdoubles/HeapArrayOfDoublesQuickSelectSketch.java
```diff
@@ -63,7 +63,8 @@
   HeapArrayOfDoublesQuickSelectSketch(final int nomEntries, final int lgResizeFactor,
       final float samplingProbability, final int numValues, final long seed) {
     super(numValues, seed);
-    nomEntries_ = nomEntries;
+    lgNomEntries_ =
+        org.apache.datasketches.Util.simpleLog2OfLong(org.apache.datasketches.Util.ceilingPowerOf2(nomEntries));
```
AlexanderSaydakov, line 67: perhaps this fully qualified name is avoidable?

leerho: I will fix using import static ...method.

PR #381, opened 2022-01-11
Fixes problems found when running latest version of SpotBugs.

leerho:
The latest version of SpotBugs is much more aggressive especially in finding potential security issues.  Some of these were false-positives, and a number only involved test code, which I'm not concerned about.  

However, there was one that did concern me of SpotBugs type "EI_EXPOSE_REP2", "This code stores a reference to an externally mutable object into the internal representation of the object."

We do this in a number of places so I was surprised that SpotBugs did not find more.  Nonetheless, the fix for this involved providing copy constructors in one hierarchy of tuple sketches.  Although I fixed the bug that SpotBugs found, I'm sure there are more.  This will require a more extensive review in the near future.

### DIFF
--- a/src/main/java/org/apache/datasketches/quantiles/DirectUpdateDoublesSketch.java
+++ b/src/main/java/org/apache/datasketches/quantiles/DirectUpdateDoublesSketch.java
@@ -253,7 +253,7 @@ final class DirectUpdateDoublesSketch extends DirectUpdateDoublesSketchR {
     memReqSvr = (memReqSvr == null) ? mem_.getMemoryRequestServer() : memReqSvr;
     if (memReqSvr == null) {
       throw new SketchesArgumentException(
-          "A reequest for more memory has been denied, "
+          "A request for more memory has been denied, "
           + "or a default MemoryRequestServer has not been provided. Must abort. ");
     }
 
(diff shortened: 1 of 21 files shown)


src/main/java/org/apache/datasketches/quantiles/DirectUpdateDoublesSketch.java
```diff
@@ -250,6 +251,11 @@ private WritableMemory growCombinedMemBuffer(final int itemSpaceNeeded) {
     assert needBytes > memBytes;
 
     memReqSvr = (memReqSvr == null) ? mem_.getMemoryRequestServer() : memReqSvr;
+    if (memReqSvr == null) {
+      throw new SketchesArgumentException(
+          "A reequest for more memory has been denied, "
```
AlexanderSaydakov, line 256: typo

leerho: Thanks!

src/main/java/org/apache/datasketches/sampling/VarOptItemsSketch.java
```diff
@@ -785,6 +785,10 @@ T getItem(final int idx) {
     return weights_.get(idx);
   }
 
+  boolean isMarksValid() {
```
AlexanderSaydakov, line 788: why expose this?

leerho: If you look at what we had before in VarOptItemsSketchTest Lines 522-528 :
```
    try {
      sketch.getMark(0);
      fail();
    } catch (final NullPointerException e) {
      // expected
    }
```
SpotBugs identified that we were catching a NullPointerException and then swallowing it.  That is a no-no.  Best practice is to check for null, but never actually "catch" a NullPointerException with a swallow.  We had a few of those.  Even though this was in test. It is still bad practice, so I fixed it.

I added isMarksValid() and changed the test to 
`assertFalse(sketch.isMarksValid());`

This avoids having to catch a NullPointerException and then swallow it.

AlexanderSaydakov (approved):

PR #389, opened 2022-02-19
Refactor kll

leerho:
Refactor KllDoublesSketch and KllFloatsSketch to extend a BaseKllSketch. 

The base class contains all the common methods that don't depend on
float or double type. This reduced a lot of duplicate code.

No substantive code changes, no changes to API.  Just moved stuff around.

### DIFF
--- a/src/main/java/org/apache/datasketches/kll/BaseKllSketch.java
+++ b/src/main/java/org/apache/datasketches/kll/BaseKllSketch.java
@@ -34,6 +34,44 @@ import org.apache.datasketches.SketchesArgumentException;
 
 abstract class BaseKllSketch {
 
+  /* Serialized float sketch layout, more than one item:
+   *  Adr:
+   *      ||    7    |   6   |    5   |    4   |    3   |    2    |    1   |      0       |
+   *  0   || unused  |   M   |--------K--------|  Flags |  FamID  | SerVer | PreambleInts |
+   *      ||   15    |   14  |   13   |   12   |   11   |   10    |    9   |      8       |
+   *  1   ||---------------------------------N_LONG---------------------------------------|
+   *      ||   23    |   22  |   21   |   20   |   19   |    18   |   17   |      16      |
+   *  2   ||<--------------data----------------| unused |numLevels|-------min K-----------|
+   *
+   *
+   *
+   * Serialized float sketch layout, Empty and Single Item:
+   *  Adr:
+   *      ||    7    |   6   |    5   |    4   |    3   |    2    |    1   |      0       |
+   *  0   || unused  |   M   |--------K--------|  Flags |  FamID  | SerVer | PreambleInts |
+   *      ||   15    |   14  |   13   |   12   |   11   |   10    |    9   |      8       |
+   *  1   ||                                   |-------------------data-------------------|
+   */
+
+  /* Serialized double sketch layout, more than one item:
+   *  Adr:
+   *      ||    7    |   6   |    5   |    4   |    3   |    2    |    1   |      0       |
+   *  0   || unused  |   M   |--------K--------|  Flags |  FamID  | SerVer | PreambleInts |
+   *      ||   15    |   14  |   13   |   12   |   11   |   10    |    9   |      8       |
+   *  1   ||---------------------------------N_LONG---------------------------------------|
+   *      ||   23    |   22  |   21   |   20   |   19   |    18   |   17   |      16      |
+   *  2   ||<-------------unused------------------------|numLevels|-------min K-----------|
+   *      ||                                                               |      24      |
+   *  3   ||<---------------------------------data----------------------------------------|
+   *
+   * Serialized double sketch layout, Empty and Single Item:
+   *  Adr:
+   *      ||    7    |   6   |    5   |    4   |    3   |    2    |    1   |      0       |
+   *  0   || unused  |   M   |--------K--------|  Flags |  FamID  | SerVer | PreambleInts |
+   *      ||                                                               |      8       |
+   *  1   ||----------------------------------data----------------------------------------|
+   */
+
   /**
    * The default value of K.
    */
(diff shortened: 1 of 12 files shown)


src/main/java/org/apache/datasketches/kll/KllDoublesSketch.java
```diff
@@ -762,65 +447,51 @@ public void merge(final KllDoublesSketch other) {
     if (Double.isNaN(maxValue_) || other.maxValue_ > maxValue_) { maxValue_ = other.maxValue_; }
     n_ = finalN;
 
-    assertCorrectTotalWeight();
+    assert KllHelper.sumTheSampleWeights(numLevels_, levels_) == n_;
```
davecromberge, line 450: I tried to find this as an invariant that was documented in the base class, but it does not appear to be listed:
```
Invariants:
 
1) After a compaction, or an update, or a merge, all levels are sorted except for level zero.
2) After a compaction, (sum of capacities) - (sum of items) >= 1, so there is room for least 1 more item in level zero.
3) There are no gaps except at the bottom, so if levels_[0] = 0, the sketch is exactly filled to capacity and must be compacted.

```

src/main/java/org/apache/datasketches/kll/KllDoublesSketch.java
```diff
@@ -287,32 +74,22 @@ public KllDoublesSketch(final int k) {
...
 
   /**
-   * Off-heap constructor.
+   * Private heapify constructor.
    * @param mem Memory object that contains data serialized by this sketch.
    */
   private KllDoublesSketch(final Memory mem) {
-    m_ = DEFAULT_M;
-    k_ = mem.getShort(K_SHORT) & 0xffff;
+    super(mem.getShort(K_SHORT) & 0xffff, DEFAULT_M, true);
```
davecromberge, line 88: I don't understand where the preamble is handled here, is it handled in the public heapify method first?

src/main/java/org/apache/datasketches/kll/KllHelper.java
```diff
@@ -65,24 +65,24 @@ static int computeTotalCapacity(final int k, final int m, final int numLevels) {
...
    * @return the capacity of a specific level
    */
-  static int levelCapacity(final int k, final int numLevels, final int height, final int minWidth) {
+  static int levelCapacity(final int k, final int numLevels, final int height, final int m) {
     assert (k <= (1 << 29));
     assert (numLevels >= 1) && (numLevels <= 61);
     assert (height >= 0) && (height < numLevels);
     final int depth = numLevels - height - 1;
-    return (int) Math.max(minWidth, intCapAux(k, depth));
+    return (int) Math.max(m, intCapAux(k, depth));
```
davecromberge, line 81: Is there a reason that you renamed `minWidth` to m, which conveys less information than the former version?

src/main/java/org/apache/datasketches/kll/BaseKllSketch.java
```diff
@@ -0,0 +1,324 @@
...
+   * The valid portion of items_ is completely packed, except for level 0.
+   * Level 0 is filled from the top down.
+   *
+   * Invariants:
+   * 1) After a compaction, or an update, or a merge, all levels are sorted except for level zero.
+   * 2) After a compaction, (sum of capacities) - (sum of items) >= 1,
+   *  so there is room for least 1 more item in level zero.
+   * 3) There are no gaps except at the bottom, so if levels_[0] = 0,
+   *  the sketch is exactly filled to capacity and must be compacted.
+   */
```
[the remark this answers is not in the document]
leerho: - An item is an abstraction of the comparable "object" that the sketch is designed to be updated with and compute the quantiles and ranks of after being updated with many such items.  For this base class a unit could be either a float or a double.
- To help the user perform _a priori_ space planning (i.e., how big will the sketch be?) , the user must start with an estimate of N, the largest anticipated number of items that the sketch will see, and the value of K to be used, which determines the error distribution of the sketch.   The user then calls the method _getMaxSerializedSizeBytes(k, N)_, which returns the maximum serialized size of the sketch.  
- Internally, in order to compute this size, the sketch first compute the upper bound on the maximum number of levels required, which is _floor(log2(N))_.  The sketch takes this information and then computes the total number of items that can be retained in a sketch given K, M, numLevels.  Where M is the internal _fixed_ value of the size of the smallest level. 
- To compute this, the sketch then iterates over all levels computing the maximum capacity of each level given K, M, numLevels.  This is a tricky computation that involves numerical folding so that the sketch could handle as many as 60 levels, which would be _extremely_ unlikely.
- The "Total Capacity" then, is the upper bound on the number of items that can be retained in the sketch given  K, M, and the maximum number of levels, which, in turn, is determined from the anticipated N. 
-  In the description, the "sum of capacities" is the "Total Capacity" given K, M, and numLevels (or N).

src/main/java/org/apache/datasketches/kll/BaseKllSketch.java
```diff
@@ -0,0 +1,324 @@
...
+  BaseKllSketch(final int k, final int m, final boolean compatible) {
+    checkK(k);
+    k_ = k;
+    minK_ = k;
+    m_ = m;
+    numLevels_ = 1;
+    levels_ = new int[] {k, k};
+    isLevelZeroSorted_ = false;
+    this.compatible = compatible;
+  }
```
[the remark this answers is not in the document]
leerho: MIN_K = DEFAULT_M = 8.  This is an artifact of the design.  In theory, this allows the possibility that in the future we might choose MIN_K > DEFAULT_M.  And the possibility that in the future, we could change the value of DEFAULT_M.  In fact, we could probably substitute DEFAULT_M for _m_ everywhere in the code because we are not allowing it to be anything else.

src/main/java/org/apache/datasketches/kll/KllDoublesSketch.java
leerho: It is an invariant as well as another invariant that I added to that text in the BaseKllSketch class.

src/main/java/org/apache/datasketches/kll/BaseKllSketch.java
```diff
@@ -0,0 +1,324 @@
...
+    while (true) {
+      assert level < numLevels_;
+      final int pop = levels_[level + 1] - levels_[level];
+      final int cap = KllHelper.levelCapacity(k_, numLevels_, level, m_);
+      if (pop >= cap) {
+        return level;
+      }
+      level++;
+    }
+  }
```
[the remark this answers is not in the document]
leerho: Yes.  If you study the compaction algorithm (designed by Kevin Lang), it is quite clever. You can use the test called _visualCheck()_ in the MiscDoublesTest class, and/or _toString(true,true)_ to help you understand how it works.

src/main/java/org/apache/datasketches/kll/KllDoublesSketch.java
leerho: The public static heapify does all the checking of the PreambleInts, SerVer, FamID and Flags.  Once these are determined to be correct, the private constructor, KllDoublesSketch(mem) does the work of populating the sketch on the heap.  The sketch in memory (heap) doesn't care about the PreambleInts, SerVer, FamID fields as those are only used to validate the serialized image, which can be stored anywhere (disk, off-heap, etc.).  

The key non-static variables of the sketch are initialized by the BaseKllSketch constructor, and then changed as necessary by the remainder of the private KllDoublesSketch(mem) construtor.

src/main/java/org/apache/datasketches/kll/KllHelper.java
leerho: The variable "m" is used everywhere else in the sketch and is defined multiples places as well.  I see no need to have another variable name that means the same thing, which adds confusion.  (How is minWidth different from m?) The answer is that it is not!

leerho:
These comments were very helpful.  I made a few minor code comment changes to improve the documentation. No code changes.

src/main/java/org/apache/datasketches/kll/BaseKllSketch.java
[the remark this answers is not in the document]
davecromberge: Thanks @leerho - the extra detail that you provided is helpful and answers my question.

src/main/java/org/apache/datasketches/kll/KllDoublesSketch.java
davecromberge: This design makes a lot of sense.

leerho:
> Does the library always use concrete primitives instead of generics because of the performance overhead for using generics? Would this also be the case with C++?

Current versions of Java Generics require generic variables to be Objects.  Boxing of primitives is space and time inefficient.  Serializing Generic objects is also inefficient.  This is why we try not to use them.   Nonetheless, we do have a few sketches that have also been implemented as Generic classes for those cases where the standard primitives are too limiting.   But these sketches cannot operate off heap.

C++ does not have any of these limitations.

leerho:
The Java Valhalla project hopes to eliminate a number of these restrictions.

PR #396, opened 2022-04-28
Cleanup warnings

leerho:
This is a lot of files but THERE ARE NO CODE CHANGES!  

This is a cleanup of unnecessary SuppressWarnings  across the whole datasketches-java library.

The vast majority of these warnings were unnecessary SuppressWarnings("javadoc") which were placed at the top of every test file. We don't require javadocs of our test files and they should never be published. This kind of suppression should be done at a global level and not in each file, which just added clutter to the code.

There were also a few Checkstyle warnings of extra empty lines.  

I wanted to clean this up before we start working on the next release.


davecromberge (approved):
Suppressing warnings at the project level makes sense.

PR #400, opened 2022-05-21
Minor method renaming that propagated to a lot of places.

leerho:


src/main/java/org/apache/datasketches/Util.java
```diff
@@ -421,41 +421,38 @@ public static boolean isPowerOf2(final int v) {
...
    * Otherwise, returns the log-base2 of the given value.
-   * @param value must be a power of 2 and greater than zero.
+   * @param powerOf2 must be a power of 2 and greater than zero.
    * @param argName the argument name used in the exception if thrown.
    * @return the log-base2 of the given value
    */
-  public static int toLog2(final int value, final String argName) {
-    checkIfPowerOf2(value, argName);
-    return Integer.numberOfTrailingZeros(value);
+  public static int exactLog2(final int powerOf2, final String argName) {
```
jmalkin, line 440: Should this and the long version have a more consistent naming scheme? e.g. exactLog2OfInt() for this one?

src/main/java/org/apache/datasketches/Util.java
```diff
@@ -421,41 +421,38 @@ public static boolean isPowerOf2(final int v) {
...
    *
-   * @param x number that is greater than zero
+   * @param powerOf2 must be a power of 2 and greater than zero.
    * @return the log2 of a long that is known to be a power of 2.
    */
-  public static int simpleLog2OfLong(final long x) {
-    final int exp = Long.numberOfTrailingZeros(x);
-    if (x != 1L << exp) {
-      throw new SketchesArgumentException("Argument x must be a positive power of 2.");
+  public static int exactLog2OfLong(final long powerOf2) {
```
jmalkin, line 451: Is there a reason we don't take a String argument here?

jmalkin (changes requested):
We should also change the pom version to 4.0.0-SNAPSHOT, so that we don't forget later.

src/main/java/org/apache/datasketches/Util.java
leerho: I'm following a java convention here that you will find in java internals.  The simpler form is for the method that is used most often.  I considered changing the other method to just exactLog2(long) since they don't collide. But it is used much less often so I figured why not just leave it alone.  If you insist on consistent naming between these two I will change the other one to exactLog2(long).

src/main/java/org/apache/datasketches/Util.java
leerho: Yes!  If this is part of a larger math expression there is no string argument to pass.  The other methods with the string argument are used when the user is supplying the argument that must be a power of 2.

jmalkin: Ok, as long as there's a principle being followed that's fine.

src/main/java/org/apache/datasketches/Util.java
jmalkin: Given the reasoning, I think I support your decision to go with this approach. No change needed.

leerho:
We're going to hold off merging this until after the next minor java version release.

davecromberge:
@leerho @jmalkin the next minor java version release (3.3.0) is busy propagating to the release mirrors and this should be safe to merge.

PR #403, opened 2022-06-25
KLL inclusive ranks and quantiles

AlexanderSaydakov:


src/main/java/org/apache/datasketches/kll/KllDoublesSketch.java
```diff
@@ -289,9 +334,20 @@ public double[] getQuantiles(final double[] fractions) {
...
+  }
+
+  /**
+   * Same as {@link #getQuantiles(int, boolean) getQuantiles(int numEvenlySpaced, false)}
+   * @param numEvenlySpaced number of evenly spaced fractional ranks
+   * @return quantiles
+   */
   public double[] getQuantiles(final int numEvenlySpaced) {
     if (isEmpty()) { return null; }
     return getQuantiles(org.apache.datasketches.Util.evenlySpaced(0.0, 1.0, numEvenlySpaced));
```
jmalkin, line 353: Doesn't explicitly call getQuantiles with inclusive=false, although the effect is still the same. Shouldn't matter unless we break the API by changing the default behavior.

jmalkin (approved):

PR #405, opened 2022-07-12
Req inclusive

AlexanderSaydakov:


leerho (approved):

leerho:
After thinking about it more, I'm approving this PR as it makes sense to have this style of API consistent with all of the quantile sketches across both C++ and Java.

PR #406, opened 2022-07-13
Extend pwr2SeriesNext and powerSeriesNextDouble to use longs.

leerho:
This simple change extends the usability of these handy methods.


leerho:
I added _christ_ reviewer by mistake.  This is a tiny change and does not impact current uses of these methods.  But it is important for some additional work I am doing in the Characterization Repo. So I'm going to go ahead and push this.

PR #408, opened 2022-07-19
Quantiles sketch sorted view

leerho:
Alex is OOV for 3 weeks, so I am going to complete this work that he started.


leerho:
I have reviewed these changes and I approve.  But because I created the PR, I cannot click "approve".
The only comment is that we should also change the internal variables like "aux" to "sortedView" or "sv", but we can do that later.  Since you obviously approve, I will do the merge.

PR #412, opened 2022-08-16
Changes to classic quantiles

leerho:


leerho:
This is a big PR but it contains some important improvements.

- All of the 5 quantile sketches now have a common set of APIs. This means that the vast majority of all the methods now have the same public method and parameter names. There are a few specific methods that are unique to some of the sketches and these will remain as part of the specific sketch and not included in the APIs.
- The 5 quantile sketches are `KllFloatsSketch`, `KllDoublesSketch`, `ReqSketch` (floats), classic `DoublesSketch` and `ItemsSketch`.
- The APIs are
     - `QuantilesAPI` -- for non type specific methods
     - `QuantilesFloatsAPI` -- for floats
     - `QuantilesDoublesAPI` -- for doubles
     Note: that due to a quirk in Java's implementation of generics, it was not possible to create a "QuantilesGenericAPI". Nonetheless, I made sure that the generic methods in the ItemsSketch do have consistent Javadocs.
- In addition, all of the sketches have a companion sorted view class and all the sorted views are implementations of a new `SortedView` interface.  Each of the sorted views have a sorted view iterator, which are implementations of a new `SortedViewIterator` interface.
- All of the quantile sketches now use a common algorithm for finding quantiles and ranks. This algorithm also accomodates both INCLUSIVE and EXCLUSIVE searches. This new approach guarantees consistency, better accuracy and handles all the corner cases correctly, which the previous methods did not. This refactoring alone allowed the removal of several helper classes and over 1000 lines of code and will be easier to maintain.
- I finally decided against the convention of using "value" for numerics and "item" for generics.  It resulted in a lot of hand editing and was very error prone. Instead, I decided to use the term "quantile" everywhere for both. It is, in fact, what these sketches are about and the term we use in our website documentation and now in our javadocs.
- I have fixed all the Checkstyle and SpotBugs issues in this PR.

src/main/java/org/apache/datasketches/kll/KllDoublesSketch.java
```diff
@@ -221,23 +221,21 @@ public double[] getQuantiles(final int numEvenlySpaced, final QuantileSearchCrit
   }
 
   /**
-   * Gets the lower bound of the quantile interval in which the true quantile of the given rank
-   * exists with a confidence of at least 99%.
-   * @param rank the given normalized rank
-   * @return the lower bound of the quantile interval in which the true quantile of the given rank
-   * exists with a confidence of at least 99%. Returns NaN if the sketch is empty.
+   * {@inheritDoc}
+   * The approximate probability that the true quantile is withing the confidence interval
```
AlexanderSaydakov, line 225: typo

src/main/java/org/apache/datasketches/kll/KllDoublesSketch.java
```diff
@@ -221,23 +221,21 @@ public double[] getQuantiles(final int numEvenlySpaced, final QuantileSearchCrit
...
   }
 
   /**
-   * Gets the upper bound of the quantile interval in which the true quantile of the given rank
-   * exists with a confidence of at least 99%.
-   * @param rank the given normalized rank
-   * @return the upper bound of the quantile interval in which the true quantile of the given rank
-   * exists with a confidence of at least 99%. Returns NaN if the sketch is empty.
+   * {@inheritDoc}
+   * The approximate probability that the true quantile is withing the confidence interval
```
AlexanderSaydakov, line 235: same typo copied all over

src/main/java/org/apache/datasketches/kll/KllDoublesSketch.java
leerho: Resolved everywhere, Thank you!

src/main/java/org/apache/datasketches/kll/KllDoublesSketch.java
leerho: Resolved everywhere, Thank you!

leerho:
The getQuantile(1.0, Exclusive) now returns the top quantile retained by the sketch, as we have discussed.
The getQuantile(rank, Exclusive) scheme implements exactly what we have discussed.

src/main/java/org/apache/datasketches/DoublesSortedView.java
```diff
@@ -36,39 +38,75 @@ public interface DoublesSortedView extends SortedView {
...
+   * <p>The points in the returned array are not monotonic and represent the discrete derivative of the CDF,
+   * which is also called the Probability Mass Function (PMF). Each returned point represents the fractional
    * area of the total distribution which lies between the previous point (or zero) and the given point, which
-   * corresponds to the given splitPoint.<p>
+   * corresponds to the given splitPoint.</p>
    *
-   * @param splitPoints the given array of quantile values or splitPoints. This is a sorted, unique, monotonic array
-   * of values in the range of (minValue, maxValue). This array should not include either the minValue or the maxValue.
+   * @param splitPoints the given array of quantile values or splitPoints. This is a sorted, monotonic array of unique
+   * values in the range of (minValue, maxValue). This array should not include either the minValue or the maxValue.
```
AlexanderSaydakov, line 88: still value here. and I am not sure about "should not include". including them is not very useful, but we don't enforce anything like that

src/main/java/org/apache/datasketches/GenericInequalitySearch.java
```diff
@@ -143,7 +143,7 @@ public static <T> int find(final T[] arr, final int low, final int high, final T
       if (hi - lo <= 1) {
         return resolve(arr, lo, hi, v, crit, comparator);
       }
-      final int mid = (lo + hi) / 2;
+      final int mid = lo + (hi - lo) / 2;
```
AlexanderSaydakov, line 146: not sure why this change to equivalent but slightly more complicated form

src/main/java/org/apache/datasketches/GenericSortedView.java
```diff
@@ -31,44 +31,64 @@
...
    * the CDF of the input data stream. Therefore, each point represents the fractional density of the distribution
-   * between from zero. For example, if one of the returned values is 0.5, then the splitPoint corresponding to that
-   * value would be the median of the distribution.</p>
+   * from zero to the given point. For example, if one of the returned values is 0.5, then the splitPoint corresponding
+   * to that value would be the median of the distribution.</p>
    *
-   * <p>If <i>isCdf</i> is false, the points in the returned array are not monotonic and represent the discrete
-   * derivative of the CDF, or the Probability Mass Function (PMF). Each returned point represents the fractional
+   * @param splitPoints the given array of quantile values or splitPoints. This is a sorted, monotonic array of unique
+   * values in the range of (minValue, maxValue). This array should not include either the minValue or the maxValue.
```
AlexanderSaydakov, line 57: still value here

src/main/java/org/apache/datasketches/GenericSortedView.java
```diff
@@ -31,44 +31,64 @@
...
+   * <p>The points in the returned array are not monotonic and represent the discrete derivative of the CDF,
+   * which is also called the Probability Mass Function (PMF). Each returned point represents the fractional
    * area of the total distribution which lies between the previous point (or zero) and the given point, which
-   * corresponds to the given splitPoint.<p>
+   * corresponds to the given splitPoint.</p>
    *
-   * @param splitPoints the given array of quantile values or splitPoints. This is a sorted, unique, monotonic array
-   * of values in the range of (minValue, maxValue). This array should not include either the minValue or the maxValue.
+   * @param splitPoints the given array of quantile values or splitPoints. This is a sorted, monotonic array of unique
+   * values in the range of (minValue, maxValue). This array should not include either the minValue or the maxValue.
```
AlexanderSaydakov, line 77: still value here

src/main/java/org/apache/datasketches/InequalitySearch.java
```diff
@@ -727,7 +727,7 @@ public static int find(final double[] arr, final int low, final int high,
       if (hi - lo <= 1) {
         return crit.resolve(arr, lo, hi, v);
       }
-      final int mid = (lo + hi) / 2;
+      final int mid = lo + (hi - lo) / 2;
```
AlexanderSaydakov, line 730: equivalent

src/main/java/org/apache/datasketches/quantiles/DoublesSketchSortedView.java
```diff
@@ -165,7 +164,7 @@ private final static void populateFromDoublesSketch(
     // Don't need to sort the corresponding weights because they are all the same.
     final int numSamples = nxt;
     Arrays.sort(itemsArr, startOfBaseBufferBlock, numSamples);
-    cumWtsArr[numSamples] = 0;
+    //cumWtsArr[numSamples] = 0;
```
AlexanderSaydakov, line 167: should this be removed?

src/main/java/org/apache/datasketches/DoublesSortedView.java
leerho: Changed it to "does not need to" here and in FloatsSortedView and GenericSortedView.

src/main/java/org/apache/datasketches/GenericInequalitySearch.java
leerho: The (lo + hi) / 2 especially with integers has a risk of overflow, while the lo + (hi - lo) / 2 does not.

src/main/java/org/apache/datasketches/GenericSortedView.java
leerho: Not sure what "still value here" means but I already corrected the "should not include" as stated above.

src/main/java/org/apache/datasketches/DoublesSortedView.java
leerho: And for both getPMF() and getCDF()

src/main/java/org/apache/datasketches/GenericSortedView.java
leerho: see comment above

src/main/java/org/apache/datasketches/InequalitySearch.java
leerho: No it ain't!  See comment above.

src/main/java/org/apache/datasketches/QuantilesDoublesAPI.java
```diff
@@ -0,0 +1,325 @@
...
+
+  /**
+   * Returns the maximum quantile of the stream. This is provided for convenience, but is distinct from the largest
+   * quantile retained by the sketch algorithm.
+   *
+   * <p>If the sketch is empty this returns NaN.</p>
+   *
+   * @return the maximum quantile of the stream
+   */
+  double getMaxQuantile();
```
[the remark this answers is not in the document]
leerho: Well it removes the ambiguity of "value", because ranks are also values.  I have also revised the web documentation to reflect this, which, in my mind makes it even more clear.  From the web doc:

"Note:
The term “value” can be ambiguous because items that we stream into a sketch are values and numeric ranks are also values. To avoid this ambiguity, we will use the term “quantiles” to refer to values that are streamed into a sketch even before they have been associated with a rank."

src/main/java/org/apache/datasketches/quantiles/ItemsSketchSortedView.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Licensed to the Apache Software Foundation (ASF) under one
+       * Licensed to the Apache Software Foundation (ASF) under one
```
AlexanderSaydakov, line 2: this seems to be a mistake

src/main/java/org/apache/datasketches/quantiles/ItemsUnion.java
```diff
@@ -126,8 +137,8 @@ public void update(final ItemsSketch<T> sketchIn) {
    * @param srcMem Memory image of sketch to be merged
    * @param serDe an instance of ArrayOfItemsSerDe
    */
-  public void update(final Memory srcMem, final ArrayOfItemsSerDe<T> serDe) {
-    final ItemsSketch<T> that = ItemsSketch.getInstance(srcMem, comparator_, serDe);
+  public void union(final Memory srcMem, final ArrayOfItemsSerDe<T> serDe) { //TODO rename merge
```
AlexanderSaydakov, line 140: should we act on this now?

src/main/java/org/apache/datasketches/req/FloatBuffer.java
```diff
@@ -19,13 +19,12 @@
 
 package org.apache.datasketches.req;
```
AlexanderSaydakov, line 21: it seems to me that getCountWithCriterion() method is not used anymore. it was called by getRank() which uses sorted view now

src/main/java/org/apache/datasketches/GenericInequalitySearch.java
AlexanderSaydakov: right, makes sense

src/main/java/org/apache/datasketches/GenericSortedView.java
AlexanderSaydakov: I mean the mention of minValue, maxValue

src/main/java/org/apache/datasketches/GenericSortedView.java
AlexanderSaydakov: this is generic sorted view, but minValue, maxValue

src/main/java/org/apache/datasketches/InequalitySearch.java
AlexanderSaydakov: right

src/main/java/org/apache/datasketches/QuantilesDoublesAPI.java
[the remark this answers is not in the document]
AlexanderSaydakov: I don't quite like the notion of the input domain consisting of quantiles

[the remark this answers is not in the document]
leerho: Do you have a better suggestion?

src/main/java/org/apache/datasketches/quantiles/DoublesSketchSortedView.java
leerho: Yes. It is now removed.

src/main/java/org/apache/datasketches/quantiles/ItemsSketchSortedView.java
leerho: Fixed. Thanks!

src/main/java/org/apache/datasketches/quantiles/ItemsUnion.java
leerho: It will make it consistent with our theta and tuple sketches.  Why not?

src/main/java/org/apache/datasketches/req/FloatBuffer.java
leerho: You are correct.  I will remove it and check the other methods in this class.

src/main/java/org/apache/datasketches/quantiles/ItemsUnion.java
AlexanderSaydakov: no objections. I just saw this comment and thought that the time for breaking changes is now.

src/main/java/org/apache/datasketches/QuantilesDoublesAPI.java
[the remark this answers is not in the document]
AlexanderSaydakov: I would say that "item" is fine generic term for input data of an arbitrary type. At least I find it less awkward than calling all input pieces of data "quantiles".

src/main/java/org/apache/datasketches/req/FloatBuffer.java
leerho: I changed my mind.  This "getCountWithCriterion()" method is used in a number of places in the ReqFloatBufferTest and removing it would also remove a number of other important tests of the FB.   Being able to test the FloatBuffer separately from the rest of the ReqSketch is valuable as it gives us confidence that the FB is operating correctly.  

The ReqSketch is a pretty complicated sketch so I am reluctant to remove code that could help us debug problems in the future.

src/main/java/org/apache/datasketches/GenericSortedView.java
leerho: Oh!  You mean the use of the word "value".  I'm in the process of fixing.

src/main/java/org/apache/datasketches/GenericSortedView.java
leerho: I get it.  I'm in the process of fixing.

AlexanderSaydakov:
well, I see that you went even further towards calling everything a quantile. I am not sure about this. I would like to get more opinions from the community.

src/main/java/org/apache/datasketches/QuantilesDoublesAPI.java
```diff
@@ -21,121 +21,125 @@
...
    *
    * <p>If the sketch is empty this returns null.</p>
    *
-   * @param splitPoints an array of <i>m</i> unique, monotonically increasing quantiles
-   * that divide the real number line into <i>m+1</i> consecutive non-overlapping intervals.
-   * It is not necessary to include either the minimum or maximum quantiles
-   * of the input stream in these split points.
+   * @param splitPoints an array of <i>m</i> unique, monotonically increasing items
+   * (of the same type as the input items)
+   * that divide the real number line into <i>m+1</i> consecutive, non-overlapping intervals.
```
AlexanderSaydakov, line 117: perhaps "input domain" instead of "real number line"

leerho: Fixed.

src/main/java/org/apache/datasketches/req/FloatBuffer.java
leerho: Keeping this method for now.

src/main/java/org/apache/datasketches/QuantilesDoublesAPI.java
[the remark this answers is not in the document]
leerho: Now, for all of the Quantile sketches, SortedViews, SortedViewIterators, SketchIterators, the following method naming is used, where applicable:

> getMaxItem()
> getMinItem()
> getRank(quantile) //because it is the inverse of getQuantile(rank)

In addition, for the code & javadocs I tried to keep the naming consistent with the above.

src/main/java/org/apache/datasketches/quantilescommon/QuantilesGenericAPI.java
```diff
@@ -0,0 +1,349 @@
...
+   * @return the normalized rank corresponding to the given quantile
+   */
+  default double getRank(T quantile) {
+    return getRank(quantile, INCLUSIVE);
+  }
+
+  /**
+   * Gets the normalized rank corresponding to the given a quantile.
+   *
+   * <p>If the sketch is empty this returns null.</p>
```
AlexanderSaydakov, line 282: NaN?

src/main/java/org/apache/datasketches/quantilescommon/QuantilesFloatsAPI.java
```diff
@@ -0,0 +1,351 @@
...
+   * <p>If the sketch is empty this returns NaN.</p>
+   *
+   * @return the minimum item of the stream
+   */
+  float getMinItem();
+
+  /**
+   * Same as {@link #getPMF(float[], QuantileSearchCriteria) getPMF(splitPoints, INCLUSIVE)}
+   * @param splitPoints an array of <i>m</i> unique, monotonically increasing items.
+   * @return a PDF array of m+1 probability masses as doubles on the interval [0.0, 1.0).
```
AlexanderSaydakov, line 99: PMF

src/main/java/org/apache/datasketches/quantiles/ItemsSketch.java
```diff
@@ -237,311 +258,369 @@ static <T> ItemsSketch<T> copy(final ItemsSketch<T> sketch) {
...
+   * @return the minimum quantile of the stream
    */
-  public T getQuantileLowerBound(final double fraction) {
-    return getQuantile(max(0, fraction - Util.getNormalizedRankError(k_, false)));
+  public T getMinItem() { return minItem_; }
+
+  /**
+   * Same as {@link #getPMF(Object[], QuantileSearchCriteria) getPMF(splitPoints, INCLUSIVE)}
+   * @param splitPoints an array of <i>m</i> unique, monotonically increasing quantiles
+   * @return a PDF array of m+1 probability masses as doubles on the interval [0.0, 1.0).
```
AlexanderSaydakov, line 339: PMF

src/main/java/org/apache/datasketches/quantiles/ItemsSketch.java
```diff
@@ -237,311 +258,369 @@ static <T> ItemsSketch<T> copy(final ItemsSketch<T> sketch) {
...
+   * @param searchCrit if INCLUSIVE, the weight of a given splitPoint quantile,
+   * if it also exists as retained quantile by the sketch,
+   * is included into the interval below.
+   * Otherwise, it is included into the interval above.
    *
-   * @param inclusive if true, the given fractional ranks are considered inclusive
-   * 
-   * @return array of approximate quantiles of the given fRanks in the same order as in the given
-   * fRanks array.
+   * @return a PDF array of m+1 probability masses as doubles on the interval [0.0, 1.0).
```
AlexanderSaydakov, line 380: PMF

src/main/java/org/apache/datasketches/quantiles/ItemsSketch.java
```diff
@@ -237,311 +258,369 @@ static <T> ItemsSketch<T> copy(final ItemsSketch<T> sketch) {
...
+   * given a set of splitPoints.
+   * The sum of the probability masses in the returned array is always 1.0.
+   *
+   * <p>The resulting approximations have a probabilistic guarantee that can be obtained from the
+   * getNormalizedRankError(true) function.
+   *
+   * <p>If the sketch is empty this returns null.</p>
+   *
+   * @param splitPoints an array of <i>m</i> unique, monotonically increasing quantiles
+   * that divide the real number line into <i>m+1</i> consecutive non-overlapping intervals.
```
AlexanderSaydakov, line 357: this is generic sketch, so perhaps "input domain" or something of that sort instead of "real number line"

src/main/java/org/apache/datasketches/quantiles/ItemsSketch.java
```diff
@@ -237,311 +258,369 @@ static <T> ItemsSketch<T> copy(final ItemsSketch<T> sketch) {
...
-   * different fractional ranks, but would be very inefficient.
-   * This method incurs the internal set-up overhead once and obtains multiple quantile values in
-   * a single query. It is strongly recommend that this method be used instead of multiple calls
-   * to getQuantile().
+   * <p>If searchCrit is EXCLUSIVE, the definition of an "interval" is
+   * inclusive of the left splitPoint and
+   * exclusive of the right splitPoint.</p>
    *
-   * <p>If the sketch is empty this returns null.
+   *<p>The left "splitPoint" for the lowest interval is the
```
AlexanderSaydakov, line 369: I don't understand this paragraph. The user provides split points, and we say just above that it is not necessary to include min and max.

src/main/java/org/apache/datasketches/quantilescommon/QuantilesGenericAPI.java
leerho: You are right null makes no sense.  But if the sketch is empty don't you think it should return zero ?  i.e., no mass?

src/main/java/org/apache/datasketches/quantilescommon/QuantilesFloatsAPI.java
leerho: Missed that, thank you!

src/main/java/org/apache/datasketches/quantiles/ItemsSketch.java
leerho: Missed that, thank you!

src/main/java/org/apache/datasketches/quantiles/ItemsSketch.java
leerho: My mistake.  This Javadoc for getPMF(splitPoints, searchCrit) was supposed to be a near exact copy of the same Javadoc for this same methods for floats and doubles, but clearly it didn't get updated.  I will fix.

leerho:
The Javadocs for the these APIs: QuantilesFloatsSketch, QuantilesDoublesSketch and QuantilesGenericSketch and the parallel methods in ItemsSketch are all supposed to be almost identical.   I will double check to make sure that they are.

Thanks.

src/main/java/org/apache/datasketches/quantilescommon/QuantilesGenericAPI.java
AlexanderSaydakov: not sure about that. if the sketch is empty then nothing is known about the distribution. it is a matter of convention I suppose

src/main/java/org/apache/datasketches/quantiles/ItemsSketch.java
AlexanderSaydakov: I would not understand this paragraph with respect to either floats or doubles

src/main/java/org/apache/datasketches/QuantilesDoublesAPI.java
leerho: This is now fixed.

AlexanderSaydakov (approved):

src/main/java/org/apache/datasketches/quantiles/ItemsSketch.java
leerho: This is now fixed.

PR #415, opened 2022-10-13
Updated POM Memory dependency to 2.2.0.

leerho:
added 2 missing package-info files.


jmalkin (approved):
i feel like we could find a better name than quantilescommon -- maybe with a hyphen rather than one word? -- but these should be primarily internal-only classes so that's ok.

PR #416, opened 2022-10-13
Initially just testing against Windows.

leerho:
This will eventually become a matrix-os, but right now I just want to test against Windows.


jmalkin (approved):
Will it work? Maybe! Let's merge and find out. Seems a bit backwards, but I guess GHA is going to require that for new workflows.

PR #418, opened 2022-10-16
troubleshooting workflows

leerho:


leerho:
I upgraded these workflows from V2 to V3 and streamlined them. They are running much faster now. 
Nonetheless, I am getting warnings as follows:

> workflow: auto-jdk-matrix.yml
> jobs/build/step:
> name: Install Matrix JDK 
> uses: actions/setup-java@v3
> 
> We get 3 of these warnings:
> 
> Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

I haven't figured out how to fix it just yet.  Not urgent, but it will cause problems eventually.

jmalkin (approved):
I don't think we need -B and -q -- I'd favor removing quiet mode, and I'm not sure if we need to care about batch -- bit this is ok.

leerho:
I presume a manually triggered GitHub/actions workflow still counts as an "automated" build?

jmalkin:
Yes, it's still considered automated in this context.

leerho:
I haven't changed it to "ready" yet. I still have the -q to fix.

PR #419, opened 2022-10-19
throwing exceptions instead of returning null or Nan when sketch is empty.

leerho:
These changes rippled throughout all the quantiles sketches.

### DIFF
--- a/src/main/java/org/apache/datasketches/quantiles/HeapCompactDoublesSketch.java
+++ b/src/main/java/org/apache/datasketches/quantiles/HeapCompactDoublesSketch.java
@@ -19,6 +19,9 @@
 
 package org.apache.datasketches.quantiles;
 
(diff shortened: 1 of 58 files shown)


src/main/java/org/apache/datasketches/quantiles/HeapCompactDoublesSketch.java
```diff
@@ -50,22 +51,22 @@ final class HeapCompactDoublesSketch extends CompactDoublesSketch {
...
+  private double maxItem_;
 
   /**
    * The total count of items seen.
    */
   private long n_;
 
   /**
-   * Number of quantiles currently in base buffer.
+   * Number of item currently in base buffer.
```
AlexanderSaydakov, line 69: should be items

AlexanderSaydakov (approved):

PR #422, opened 2022-12-10
This fixes all of the "High", "Error", and "Warnings" issues detected by

leerho:
CodeQL. However, there were some that I classified as "Won't Fix" and "False Positive".  But I'm not sure if CodeQL will remember those markings.


davecromberge (approved):
CodeQL looks useful - is there still a need to retain the Checkstyle tool as part of the release?

leerho:
CodeQL and Checkstyle are still quite different with different objectives. I think we need both.

PR #427, opened 2023-02-08
Theta compact iterator

AlexanderSaydakov:
Added HeapCompactHashIterator that does not need to check hash values for 0 and less than theta.
Removed unnecessary member in HeapHashIterator.

### DIFF
--- a/src/main/java/org/apache/datasketches/theta/EmptyCompactSketch.java
+++ b/src/main/java/org/apache/datasketches/theta/EmptyCompactSketch.java
@@ -113,7 +113,7 @@ final class EmptyCompactSketch extends CompactSketch {
 
   @Override
   public HashIterator iterator() {
-    return new HeapHashIterator(new long[0], Long.MAX_VALUE);
+    return new HeapCompactHashIterator(new long[0]);
   }
 
   /**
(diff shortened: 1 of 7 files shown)


jmalkin (approved):

leerho (approved):
Since _HashIterator_ is the only one of this group that is actually returned to the user it should remain in the _theta_ package.  But the _HeapCompactHashIterator_ is used by the _EmptyCompactSketch_, _HeapCompactSketch_ and the _SingleItemSketch_, and the _HeapHashIterator_ is used by the _HeapAlphaSketch_ and the HeapQuickSelectSketch. 

I think both the _HeapCompactHashIterator_ and the _HeapHashIterator_ would be good candidates to move them to the _thetacommon_ package.

This is not a requirement, but I would like to, over time, clean up the _theta_ package to make it easier to find the classes that actually drive the API.   So far, the classes I have moved to the _thetacommon_ package were all public anyway, so it was easy.  

What has become the practice in the JVM is to have only public API classes at the root and moving all other auxiliary classes to an _internal_ package off the root.  This means these classes often have to be made _public_, but when encapsulated in a JPMS module (for java9+), the _internal_ package can be made non-visible to the user.  
 
So the other possibility would be to create _internal_ sub packages for our sketches as well.

PR #433, opened 2023-03-20
Apparently, the name of the codeQL workflow must be codeql-analysis.yml.

leerho:


leerho:
This PR just changes the name of the github/workflows/codeql.yml to github/workflows/codeql-analysis.yml, which I found out is the proper default name.  

However, there remains a codeql warning "1 configuration present on refs/heads/master was not found:", but I cannot find any useful info on line on how to correct this warning.  

It turns out that changes to workflows must exist in master first, before submitting any branch PRs that also reference these workflows, so I am pushing this single name change without review.

PR #440, opened 2023-05-03
This commit fixes all the SpotBugs 4.7.3 issues.

leerho:
I also cleaned up the tools/FindBugsExcludeFilter.xml.

Jon & I fixed some actual bugs that were uncovered by SB, and came up with superior solutions.  These were mostly in quantiles, sampling and VarOpt where we were not checking the possibility of improper method call sequences corrupting the sketch.  (long story).

In addition, Will & I came up with an elegant solution to being able to use SpotBugs and SpotBugs @SuppressFBWarnings WITHOUT having to make SpotBugs a dependency. It works GREAT!

### DIFF
--- a/src/main/java/org/apache/datasketches/quantiles/ItemsSketchIterator.java
+++ b/src/main/java/org/apache/datasketches/quantiles/ItemsSketchIterator.java
@@ -19,6 +19,9 @@
 
 package org.apache.datasketches.quantiles;
 
+import java.util.Objects;
+
+import org.apache.datasketches.common.SketchesStateException;
 import org.apache.datasketches.quantilescommon.QuantilesGenericSketchIterator;
 
 /**
@@ -26,8 +29,7 @@ import org.apache.datasketches.quantilescommon.QuantilesGenericSketchIterator;
  * @param <T> type of item
  */
 public class ItemsSketchIterator<T> implements QuantilesGenericSketchIterator<T> {
-
-  private final ItemsSketch<T> sketch;
+  //private final ItemsSketch<T> sketch;
   private Object[] combinedBuffer;
   private long bitPattern;
   private int level;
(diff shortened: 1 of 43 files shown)


leerho:
Coverage is 97.8% for the Main code.

src/main/java/org/apache/datasketches/quantiles/ItemsSketchIterator.java
```diff
@@ -19,35 +19,42 @@
...
 import org.apache.datasketches.quantilescommon.QuantilesGenericSketchIterator;
 
 /**
  * Iterator over ItemsSketch. The order is not defined.
  * @param <T> type of item
  */
 public class ItemsSketchIterator<T> implements QuantilesGenericSketchIterator<T> {
-
-  private final ItemsSketch<T> sketch;
+  //private final ItemsSketch<T> sketch;
```
jmalkin, line 32: can remove this entirely if it's unneeded

src/main/java/org/apache/datasketches/req/ReqCompactor.java
```diff
@@ -49,7 +49,6 @@ class ReqCompactor {
   //objects
   private FloatBuffer buf;
   private final ReqDebug reqDebug = null;
-  private Random rand;
```
jmalkin, line 52: I'm surprised SpotBugs complained about the use of random here

jmalkin (approved):
Approving, although someone who better understands the FindBugsExcludeFilter.xml changes is encouraged.

src/main/java/org/apache/datasketches/req/ReqCompactor.java
leerho: Actually, it was complaining that an instance of Random was created and then used only once.  This is not only slow, but it does not produce good random behavior.  

So I took a hard look at how Random was used in REQ and refactored the use of Random in REQ completely.  Now, Random is instantiated once in ReqSketch and then a reference to it is passed to the ReqCompactor.compact(...) method which then uses the same random instance for every compaction.  This also opens the door for us to easily add a "makeDeterministic" flag to the public constructor if we decide we want to do that (which I am increasing thinking that this would be ok to do).

src/main/java/org/apache/datasketches/quantiles/ItemsSketchIterator.java
leerho: I removed this comment and will bundle this fix with changes to the HLL Javadocs.

PR #447, opened 2023-05-30
Interim 1 towards KllItemsSketch - works

leerho:


src/main/java/org/apache/datasketches/common/ArrayOfDoublesSerDe.java
```diff
@@ -29,28 +31,58 @@
...
+  @Override
+  @Deprecated
+  public Double[] deserializeFromMemory(final Memory mem, final int numItems) {
+    return deserializeFromMemory(mem, 0, numItems);
+  }
+
+  @Override
+  public Double[] deserializeFromMemory(final Memory mem, final long offsetBytes, final int numItems) {
+    long offset = offsetBytes;
+    Util.checkBounds(offset, Double.BYTES * numItems, mem.getCapacity());
```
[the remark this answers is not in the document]
leerho: All fixed.

src/main/java/org/apache/datasketches/common/ArrayOfLongsSerDe.java
```diff
@@ -29,28 +31,57 @@
...
+  @Override
+  @Deprecated
+  public Long[] deserializeFromMemory(final Memory mem, final int numItems) {
+    return deserializeFromMemory(mem, 0, numItems);
+  }
+
+  @Override
+  public Long[] deserializeFromMemory(final Memory mem, final long offsetBytes, final int numItems) {
+    long offset = offsetBytes;
+    Util.checkBounds(offset, Long.BYTES * numItems, mem.getCapacity());
```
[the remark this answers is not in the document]
leerho: All fixed.

src/main/java/org/apache/datasketches/common/ArrayOfNumbersSerDe.java
```diff
@@ -144,7 +127,76 @@
               + i + ": " + numType);
       }
     }
-
     return array;
   }
+
+  @Override
+  public int sizeOf(final Number item) {
+    if ( item instanceof Long)         { return Byte.BYTES + Long.BYTES; }
```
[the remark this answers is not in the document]
leerho: All fixed.

src/main/java/org/apache/datasketches/common/ArrayOfNumbersSerDe.java
```diff
@@ -45,106 +51,218 @@
   private static final byte DOUBLE_INDICATOR  = 4;
   private static final byte FLOAT_INDICATOR   = 6;
 
+  @Override
+  public byte[] serializeToByteArray(final Number item) {
+    final byte[] byteArr;
+    if (item instanceof Long) {
```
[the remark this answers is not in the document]
leerho: All fixed

leerho:
I might suggest that instead of starting with the GitHub review tool, that you first acquaint yourself with the overall structure of the entire KLL code.  I have structured both the main code and the test classes so that the floats classes are nearly identical to the doubles classes, line for line, with the exception of where the data type is specified.  The items classes are also very similar with the key exception that with Java 8 technology it is not feasible to create a writable off-heap implementation.  When we get to Java 21, this will become more doable.  The similarity between classes of the same name, except for the type, should simplify the review process.

src/main/java/org/apache/datasketches/sampling/ReservoirItemsSketch.java
```diff
@@ -259,7 +259,7 @@ public static <T> ReservoirItemsSketch<T> heapify(final Memory srcMem,
 
     final int itemsToRead = (int) Math.min(k, itemsSeen);
     final T[] data = serDe.deserializeFromMemory(
-        srcMem.region(preLongBytes, srcMem.getCapacity() - preLongBytes), itemsToRead);
+        srcMem.region(preLongBytes, srcMem.getCapacity() - preLongBytes), 0, itemsToRead);
```
AlexanderSaydakov, line 262: It is not clear to me why this region is needed here. this new offset parameter can be used instead. or, the other way around, why is this new offset parameter needed, if we always can use a region?

src/main/java/org/apache/datasketches/common/ArrayOfStringsSerDe.java
```diff
@@ -31,45 +34,101 @@
...
 
+  @Override
+  public int sizeOf(final String item) {
+    Objects.requireNonNull(item, "Item must not be null");
+    if (item.isEmpty()) { return Integer.BYTES; }
+    return item.getBytes(StandardCharsets.UTF_8).length + Integer.BYTES;
+  }
+
+  @Override
+  public int sizeOf(final Memory mem, final long offsetBytes, final int numItems) {
```
AlexanderSaydakov, line 111: what is the purpose of this? seems like dry-run deserialization

src/main/java/org/apache/datasketches/kll/KllDirectCompactItemsSketch.java
```diff
@@ -0,0 +1,307 @@
...
+  @Override
+  byte[] getMinMaxByteArr() {
+    if (sketchStructure == COMPACT_EMPTY || getN() == 0) { return new byte[0]; }
+    if (sketchStructure == COMPACT_SINGLE) {
+      final int bytesSingle = serDe.sizeOf(mem, DATA_START_ADR_SINGLE_ITEM, 1);
+      final byte[] byteArr = new byte[2 * bytesSingle];
+      mem.getByteArray(DATA_START_ADR_SINGLE_ITEM, byteArr, 0, bytesSingle);
+      mem.getByteArray(DATA_START_ADR_SINGLE_ITEM, byteArr, bytesSingle, bytesSingle);
+    }
+    //sketchStructure == COMPACT_FULL
```
AlexanderSaydakov, line 140: seems like COMPACT_SINGLE also reaches this point

src/main/java/org/apache/datasketches/kll/KllDirectCompactItemsSketch.java
```diff
@@ -0,0 +1,307 @@
...
+  @Override
+  T[] getRetainedItemsArray() {
+    final int numRet = getNumRetained();
+    final T[] outArr = (T[]) Array.newInstance(serDe.getClassOfT(), numRet);
+    if (sketchStructure == COMPACT_EMPTY) { return outArr; }
+    if (sketchStructure == COMPACT_SINGLE) {
+      outArr[0] = getSingleItem();
+      return outArr;
+    }
+    final int offset = getCompactDataOffset(); //both single & full
```
AlexanderSaydakov, line 168: single doesn't seem to reach this point

src/main/java/org/apache/datasketches/kll/KllHeapItemsSketch.java
```diff
@@ -0,0 +1,297 @@
...
+
+  @Override
+  public int getK() {
+    return k;
+  }
+
+  @Override
+  public T getMaxItem() {
+    if (isEmpty()) { throw new SketchesArgumentException(EMPTY_MSG); }
+    return (T)maxItem;
```
AlexanderSaydakov, line 129: maxItem is of type T already

src/main/java/org/apache/datasketches/kll/KllHeapItemsSketch.java
```diff
@@ -0,0 +1,297 @@
...
+  }
+
+  @Override
+  public T getMinItem() {
+    if (isEmpty()) { throw new SketchesArgumentException(EMPTY_MSG); }
+    return (T)minItem;
```
AlexanderSaydakov, line 135: minItem is of type T already

src/main/java/org/apache/datasketches/common/ArrayOfStringsSerDe.java
leerho: We do not know whether T is variable in size or not, only the user knows.  If T happens to be fixed in size, the computation of the size of the array of items is a trivial multiply, which is far less expensive than a deserialization.  If T varies in size than the user will have to scan the in-memory array and determine its size.  Not having this method would mean forcing a deserialization and then grabbing the size of the array even when a trivial multiply would do.

src/main/java/org/apache/datasketches/kll/KllDirectCompactItemsSketch.java
leerho: This actually works because `getSingleItem(`) does the work. However, this method would be easier to read if I had removed the COMPACT_SINGLE case like I did with `getRetainedItemsByteArr()` and `getRetainedItemsSizeBytes()`, because the `getCompactDataOffset()` computes the correct offset for both Compact Single and Full cases.

src/main/java/org/apache/datasketches/kll/KllDirectCompactItemsSketch.java
leerho: Very good catch!  Thank you! Fixed.

src/main/java/org/apache/datasketches/kll/KllHeapItemsSketch.java
leerho: Fixed.

src/main/java/org/apache/datasketches/kll/KllHeapItemsSketch.java
leerho: Fixed

leerho:
This code has not been tested against files generated by C++.  I am sure there will be issues.

src/main/java/org/apache/datasketches/sampling/ReservoirItemsSketch.java
leerho: You are correct.  My modification only made the error go away, I should have replaced the use of region entirely.  Fixed.

src/main/java/org/apache/datasketches/kll/KllDirectCompactItemsSketch.java
leerho: Fixed.

src/test/java/org/apache/datasketches/kll/KllFloatsSketchSerDeTest.java
```diff
@@ -64,7 +64,7 @@ public void serializeDeserializeOneValue() {
     assertEquals(sketch2.getNormalizedRankError(false), sketch1.getNormalizedRankError(false));
     assertEquals(sketch2.getMinItem(), 1.0F);
     assertEquals(sketch2.getMaxItem(), 1.0F);
-    assertEquals(sketch2.getSerializedSizeBytes(), 8 + Float.BYTES);
+    assertEquals(sketch2.getSerializedSizeBytes(), sketch2.getSerializedSizeBytes());
```
AlexanderSaydakov, line 67: this is always true. perhaps something else was meant?

src/test/java/org/apache/datasketches/kll/KllItemsSketchSerDeTest.java
```diff
@@ -0,0 +1,105 @@
...
+    final byte[] bytes = sketch1.toByteArray();
+    final KllItemsSketch<String> sketch2 = KllItemsSketch.heapify(Memory.wrap(bytes), Comparator.naturalOrder(), serDe);
+    assertEquals(bytes.length, sketch1.getSerializedSizeBytes());
+    assertFalse(sketch2.isEmpty());
+    assertEquals(sketch2.getNumRetained(), 1);
+    assertEquals(sketch2.getN(), 1);
+    assertEquals(sketch2.getNormalizedRankError(false), sketch1.getNormalizedRankError(false));
+    assertEquals(sketch2.getMinItem(), " 1");
+    assertEquals(sketch2.getMaxItem(), " 1");
+    assertEquals(sketch2.getSerializedSizeBytes(), sketch2.getSerializedSizeBytes());
```
AlexanderSaydakov, line 68: should this compare sketch1 and sketch2?

src/test/java/org/apache/datasketches/kll/KllItemsSketchSerDeTest.java
```diff
@@ -0,0 +1,105 @@
...
+  //no cross language tests yet
+
+  @Test(groups = {"generate"})
+  public void generateBinariesForCompatibilityTesting() throws Exception {
+    final int[] nArr = {0, 1, 10, 100, 1_000, 10_000, 100_000, 1_000_000};
+    for (int n: nArr) {
+      final int digits = numDigits(n);
+      final KllItemsSketch<String> sketch = KllItemsSketch.newHeapInstance(Comparator.naturalOrder(), serDe);
+      for (int i = 0; i < n; i++) sketch.update(intToFixedLengthString(i, digits));
+      try (final FileOutputStream file = new FileOutputStream("kll_items_n" + n + ".sk")) {
```
AlexanderSaydakov, line 99: I would suggest naming the files "kll_string_n" to be consistent with the naming in other places

src/test/java/org/apache/datasketches/kll/KllFloatsSketchSerDeTest.java
leerho: Oops!  I should have just left this one alone.  I was trying to eliminate the magic number :(

src/test/java/org/apache/datasketches/kll/KllItemsSketchSerDeTest.java
leerho: Oops again. Fixed. (will post soon)
Nonetheless, I have extended these tests for Floats, Doubles, and Items so that they also test SerDe round trip between heap and offheap.

src/test/java/org/apache/datasketches/kll/KllItemsSketchSerDeTest.java
leerho: Done

src/main/java/org/apache/datasketches/kll/KllItemsSketch.java
```diff
@@ -0,0 +1,366 @@
...
+  public final void merge(final KllSketch other) {
+    if (readOnly || sketchStructure != UPDATABLE) { throw new SketchesArgumentException(TGT_IS_READ_ONLY_MSG); }
+    final KllItemsSketch<T> othItmSk = (KllItemsSketch<T>)other;
+    if (othItmSk.isEmpty()) { return; }
+    KllItemsHelper.mergeItemImpl(this, othItmSk, comparator);
+    kllItemsSV = null;
+  }
+
+  @Override
+  public void reset() {
```
AlexanderSaydakov, line 264: perhaps missing reset of sorted view

src/main/java/org/apache/datasketches/kll/KllMemoryValidate.java
```diff
@@ -45,157 +32,162 @@
...
 
-    switch (sw) {
-      case 0: { //FULL_COMPACT
-        if (preInts != PREAMBLE_INTS_FULL) { memoryValidateThrow(INVALID_PREINTS, preInts); }
-        if (serVer != SERIAL_VERSION_EMPTY_FULL) { memoryValidateThrow(EMPTYBIT_AND_SER_VER, serVer); }
+    switch (sketchStructure) {
+      case COMPACT_FULL: {
+        if (emptyFlag) { throw new SketchesArgumentException(EMPTY_FLAG_AND_COMPACT_FULL); }
         n = getMemoryN(srcMem);
+        //if (n <= 1) { memoryValidateThrow(N_AND_COMPACT_FULL); }
```
AlexanderSaydakov, line 118: do we need ti keep this commented-out line?

src/main/java/org/apache/datasketches/kll/KllItemsSketch.java
leerho: Thank you for catching this!  
It is now fixed.  see the push notes.

src/main/java/org/apache/datasketches/kll/KllMemoryValidate.java
leerho: Ah, this was a result of a discussion we had WRT an "improperly" compacted sketch.  This is where a sketch has only one value, but was not compacted using the Single Item serialization layout.  I think we concluded that this could happen with a very old sketch built prior to the introduction of the Single Item layout.  

I added a comment to that effect so that we can remember why we can't use this test.  I also commented out
the associated error message string at the bottom of the class.

AlexanderSaydakov (approved):

PR #450, opened 2023-06-07
Preparatory work for generic items.

leerho:
1. The UpdatableBitMask as part of the Flags field has been eliminated. This should make the Flags field identical to the one used in C++.

2. The documentation of the serialization formats has been significantly improved.  See the docs for the KllPreambleUtil class.

3. I have reduced the dependence on the KllMemoryValidate class to those that are actually required for validation.

4. More cleanup of fields and variables not really being used.


leerho:
I forgot to mention, this also does some preparation of the current master for a ds-java 4.1.0 release.

davecromberge (approved):

PR #453, opened 2023-06-13
Update theta compact() behavior.

leerho:
Update associated javadocs.

### DIFF
--- a/src/main/java/org/apache/datasketches/theta/Sketch.java
+++ b/src/main/java/org/apache/datasketches/theta/Sketch.java
@@ -39,7 +39,7 @@ import org.apache.datasketches.thetacommon.BinomialBoundsN;
 import org.apache.datasketches.thetacommon.ThetaUtil;
 
 /**
- * The top-level class for all sketches. This class is never constructed directly.
+ * The top-level class for all theta sketches. This class is never constructed directly.
  * Use the UpdateSketch.builder() methods to create UpdateSketches.
  *
  * @author Lee Rhodes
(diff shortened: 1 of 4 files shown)


leerho:
Will noticed that when calling compact() on a Theta CompactSketch, instead of just returning `this` the code created another compact object.  After digging into it, I realized that not only was the code not optimal, it was not well documented either.  The test code also needed to be rewritten to reflect this updated behavior. This PR fixes these issues.

src/test/java/org/apache/datasketches/theta/CompactSketchTest.java
```diff
@@ -220,81 +221,226 @@ public void checkMemTooSmallOrdered() {
 
   @Test
   public void checkCompactCachePart() {
-    //phony values except for curCount = 0.
+    //phoney values except for curCount = 0.
```
jmalkin, line 224: We can switch to UK spellings but why only this one place in an internal comment?

src/main/java/org/apache/datasketches/theta/Sketch.java
```diff
@@ -193,43 +193,50 @@ public static Sketch wrap(final Memory srcMem, final long expectedSeed) {
   //Sketch interface
 
   /**
-   * Converts this sketch to a ordered CompactSketch on the Java heap.
+   * Converts this sketch to a ordered CompactSketch.
    *
-   * <p>If this sketch is already in the proper form, this method returns <i>this</i>,
-   * otherwise, this method returns a new CompactSketch of the proper form.
+   * <p>If <i>this.isCompact() == true</i> this method returns <i>this</i>,
```
jmalkin, line 198: Two things here.
1) I think <em> is generally preferred to <i> in that it's semantic rather than direct typesetting. What does italics mean if using a screen reader, for instance?
2) For code snippets that aren't put into their own preformatted block, I think <tt> is generally preferred over italicized text.

That's going to apply to a _lot_ of code in this repo, though, so perhaps better suited to a dedicated PR that does only that change.

jmalkin:
> Do we have to have compact() method in the compact theta sketch class? I would think it should exist in the updatable sketch only.

`compact()` is defined in the top-level Sketch interface so it'd be a major version bump to change that.

src/main/java/org/apache/datasketches/theta/Sketch.java
leerho: I'm willing to change it **here** but your comments don't give me a clue whether to use `<em>` or `<tt>`.

src/test/java/org/apache/datasketches/theta/CompactSketchTest.java
leerho: Funny, when I did my spellcheck it suggested "phoney"  I didn't give it a second glance and realize that this is the British spelling :(

src/main/java/org/apache/datasketches/theta/Sketch.java
jmalkin: Hoping I can get a comment from someone else. I think I'd vote for `<tt>` -- but I was more using it as a discussion point since I noticed here. If we're going to change the style we should try to do this everywhere in a single pass, IMHO. Or at least a series of PRs per sketch type/family.

jmalkin: Oops.. ok, so then we should use the `<code>` replacement :)

leerho: It is not so clear that \<em\> is necessarily the correct answer either.  Even though Java documentation does use it.  

"The \<i\> element represents text that is set off from the normal prose, such as a foreign word, fictional character thoughts, or when the text refers to the definition of a word instead of representing its semantic meaning." 

"An example for \<em\> could be: "Just _do_ it already!", or: "We _had_ to do something about it". A person or software reading the text would pronounce the words in italics with an emphasis, using verbal stress." [The emphasis element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/em)

I don't think someone reading these javadoc aloud would use verbal stress on these words. 

Is it such a big deal?

leerho: The [Oracle Style Guide](https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#styleguide) for Javadocs recommends using \<code\>...\</code\> for 

- Java keywords
- package names
- class names
- method names
- interface names
- field names
- argument names
- code examples

This would represent a big cleanup of the entire repo if we were to decide to do this.

leerho:
What changes are required to approve this?  So far, I'll agree to the spelling change.

src/main/java/org/apache/datasketches/theta/Sketch.java
jmalkin: Yes, we should definitely make that change. But as I said earlier, I think it makes more sense to do it as dedicated PRs per sketch family.

I'll resolve this for now since the concept has now been discussed.

will-lauer: I know I'm late to this converstation, but don't forget about the `{@code}` annotation in javadoc. This is also a possibility instead of `<code>` and definitely preferable to things like `<em>` and `<i>`

will-lauer (approved):
Looks good to me, other than the use of, `<i>` vs `<code>` or `{@code}`, but I understand that's a bigger cleanup and not part of this PR.

PR #464, opened 2023-08-07
Test SerDe compatibility with C++

AlexanderSaydakov:


AlexanderSaydakov:
Also changed the EMPTY_FLAG_MASK in frequent items sketch because of a mistake in C++, where a different bit was used. 
The solution is to set and check both bits since no other bits in the flags byte were used.

PR #465, opened 2023-08-15
Interim 1 Cross lang testing

leerho:
Fixed KllItems generation starting with 1


leerho:
This completes 1st phase changes to support cross-language testing between C++ and Java.  

There are two new profiles created in the pom.xml:  

- mvn clean test -P generate-java-files

This runs a TestNG Group that generates pre-defined serialized sketches into the directory "/target/java_generated_files/".  

- mvn clean test -P check-cpp-files

This runs a TestNG Group that runs tests on files currently saved in the directory "/src/test/resources/".

These directories are set by static final constants in 
org.apache.datasketches.common.TestUtil and can easily be changed.

The old C++ files that are in "/src/test/resources/" will be eventually deleted when we have a new set of C++ files.
At that time we should change the path for these files to live under the target directory like the java files.  
My recommended local path for that directory is "/target/cpp_generated_files".
We will also have to create the Java tests that test those files as well as the script to obtain them.

leerho:
I split out the historical files into a separate profile with its potentially separate storage path, but currently still in "src/test/resources"

The current ones are also still in "src/test/resources" until we decide to move them to a permanent path.   

Links to look at:
[TestUtil](https://github.com/apache/datasketches-java/blob/kll_cross_lang_IT/src/test/java/org/apache/datasketches/common/TestUtil.java)
[pom.xml](https://github.com/apache/datasketches-java/blob/kll_cross_lang_IT/pom.xml); L91-94,  L290, L685, L703-704, L722-723, L741-742.

This might also help:
`GROUP NAME                     POM TAG                             POM Profile
"check_cpp_files"               testng.check-cpp-files              check-cpp-files
"generate_java_files"           testng.generate-java-files          generate-java-files
"check_cpp_historical_files"    testng.check-cpp-historical-files   check-cpp-historical-files`

AlexanderSaydakov (approved):

PR #474, opened 2023-10-24
Revert to class loader get resource

leerho:
This PR does 2 things:

1. In the prior release I thought I didn't need the Java Class.getResource() methods for obtaining resources.  Well, I was wrong. So this reinstalls these important methods with a major improvement: These methods will now work from JAR files as well as from within a file system.

2. I increased my detection of all kinds of minor code quality issues.  So this commit fixes all that I could find. Examples: redundant casts, redundant declaration of generic parameters, methods that could be declared static, variable name hiding, etc.


leerho:
Also removed some deprecated code, which has been there since 2.0.
Moved the Cross-language target to target2/ and ignored it in gitignore.
A few other minor fixes, e.g, spelling etc.

AlexanderSaydakov (approved):

PR #479, opened 2023-12-05
Release Process: Change pom version to 5.1.0-SNAPSHOT

leerho:
Release Process: sets the pom version to 5.1.0-SNAPSHOT for main-line development.


jmalkin (approved):
Apparently we forgot to do 5.0-SNAPSHOT. I guess the major version bump wasn't initially planned :)

PR #480, opened 2023-12-05
Ability-to-update-multiple-identical-elements (focus is on the quanti…

edmondliuTTD:
We have a use case where we want to create quantile sketches and update them with the same element multiple times.


leerho:
Please see comments in [Issue 481](https://github.com/apache/datasketches-java/issues/481)
This PR is being closed

PR #482, opened 2023-12-06
HLL Union toString is not pure and can trigger a rebuild

freakyzoidberg:
Hello,

The toString method of Union can mutate the object if a rebuild is necessary.
This make some test fail when running in debug (as some IDE will render object value automatically).

Not a major issue, but surely unexpected to have toString() mutate the object.

Thanks !


leerho (approved):

leerho:
Thank you!  Excellent fix!

PR #483, opened 2023-12-07
Add final

leerho:
The last commit was missing a final.


jmalkin (approved):
short-lived enough an item that it shouldn't matter, but this seems good

PR #485, opened 2023-12-17
Remove duplicate code

leerho:
I'm trying to do the next set of updates in smaller batches.

### DIFF
--- a/src/main/java/org/apache/datasketches/kll/KllHelper.java
+++ b/src/main/java/org/apache/datasketches/kll/KllHelper.java
@@ -718,7 +718,7 @@ final class KllHelper {
     return result;
   }
 
-  private final static boolean enablePrinting = true;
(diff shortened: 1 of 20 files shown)


src/main/java/org/apache/datasketches/kll/KllHelper.java
```diff
@@ -806,31 +707,32 @@ private static long intCapAux(final int k, final int depth) {
   /**
    * Performs the integer based calculation of an individual level (or folded level).
    * @param k the configured k of the sketch
-   * @param depth depth the zero-based index of the level being computed.
+   * @param depth the zero-based index of the level being computed. The max depth is 30!
```
AlexanderSaydakov, line 710: do we need "!" in docs?

AlexanderSaydakov (approved):

PR #489, opened 2024-01-08
small changes in the update single and update weighted methods.

leerho:
This fixes a problem where the update to min and max might not occur on merging.

This also simplifies the code and the number of checks in the update path.


jmalkin (approved):
Don't think I'd realized we were farming off update to a helper class. Java seems so much more complex than C++!

Anyway, this seems fine.

PR #490, opened 2024-01-09
Changed the name of a utility function that should have been defined for longs.

leerho:
This simple changed affected 10 files, but it is just a name change of one function.


leerho:
The second function is always used together with the first function.

AlexanderSaydakov (approved):

PR #499, opened 2024-01-30
EB-PPS Sampling sketch

jmalkin:
Adds EBPPS Sketch, providing parity with Java.

No cross-language tests yet. I did perform some characterization tests for entropy of a single sketch and 2-way merge, where the average of many runs should look like the weighted input distribution and things were in line with the C++ version. I haven't figured out how to create proper tests from that since it's basically "Are entropy and KL Divergence 'close enough'?"

### DIFF
--- a/src/main/java/org/apache/datasketches/sampling/EbppsItemsSample.java
+++ b/src/main/java/org/apache/datasketches/sampling/EbppsItemsSample.java
@@ -21,6 +21,7 @@ package org.apache.datasketches.sampling;
 
 import static org.apache.datasketches.common.Util.LS;
 
+import java.lang.reflect.Array;
 import java.util.ArrayList;
 import java.util.Random;
 import java.util.concurrent.ThreadLocalRandom;
@@ -122,12 +123,30 @@ class EbppsItemsSample<T> {
     return result;      
   }
 
-  // package-private for use in merge
+  @SuppressWarnings("unchecked")
+  T[] getAllSamples(final Class<?> clazz) {
+    // Is it faster to use sublist and append 1?
+    final T[] itemsArray = (T[]) Array.newInstance(clazz, getNumRetainedItems());
+    int i = 0;
+    if (data_ != null) {
+      for (T item : data_) {
+        if (item != null) {
+          itemsArray[i++] = item;
+        }
+      }
+    }
+    if (partialItem_ != null)
+      itemsArray[i] = partialItem_; // no need to increment i again
+
+    return itemsArray;
+  }
+
+  // package-private for use in merge and serialization
   ArrayList<T> getFullItems() {
     return data_;
   }
 
-  // package-private for use in merge
+  // package-private for use in merge and serialization
   T getPartialItem() {
     return partialItem_;
   }
@@ -296,28 +315,4 @@ class EbppsItemsSample<T> {
     return (data_ != null ? data_.size() : 0)
          + (partialItem_ != null ? 1 : 0);
   }
-
-  public int getSerializedSizeBytes(final ArrayOfItemsSerDe<? super T> serDe) {
-    return -1;  
-  }
-
-  public byte[] toByteArray(final ArrayOfItemsSerDe<? super T> serDe) {
-    if (data_.size() == 0 && partialItem_ == null) {
-      // null class is ok since empty -- no need to call serDe
-      return toByteArray(serDe, null);
-    } else {
-      return toByteArray(serDe, data_.get(0).getClass());
-    }
-  }
-
-  public byte[] toByteArray(final ArrayOfItemsSerDe<? super T> serDe, final Class<?> clazz) {
-    return null;
-  }
-
-
-  public static <T> EbppsItemsSample<T> heapify(final Memory srcMem, final ArrayOfItemsSerDe<T> serDe) {
-    return new EbppsItemsSample<>(0);
-  }
-
-
 }
--- a/src/main/java/org/apache/datasketches/sampling/EbppsItemsSample.java
+++ b/src/main/java/org/apache/datasketches/sampling/EbppsItemsSample.java
@@ -97,8 +97,8 @@ class EbppsItemsSample<T> {
       return null;
 
     ArrayList<T> result = new ArrayList<>(resultSize);
-    for (T item : data_)
-      result.add(item);
+    if (data_ != null)
+      result.addAll(data_);
 
     if (includePartial)
       result.add(partialItem_);
--- a/src/main/java/org/apache/datasketches/sampling/EbppsItemsSample.java
+++ b/src/main/java/org/apache/datasketches/sampling/EbppsItemsSample.java
@@ -34,7 +34,7 @@ class EbppsItemsSample<T> {
 
   private double c_;            // Current sample size, including fractional part
   private T partialItem_;       // a sample item corresponding to a partial weight
-  private ArrayList<T> data_;   // stored sampled items
+  private ArrayList<T> data_;   // full sample items
   
   private Random rand_;         // ThreadLocalRandom.current() in general
 
@@ -52,20 +52,6 @@ class EbppsItemsSample<T> {
     data_ = new ArrayList<>(other.data_);
     rand_ = other.rand_;
   }
-
-  /*
-  // constructor used to create a sample to merge one item
-  EbppsItemsSample(final T item, final double theta) {
-    c_ = theta;
-    if (theta == 1.0) {
-      data_ = new ArrayList<T>(1);
-      data_.add(item);
-    } else {
-      partialItem_ = item;
-    }
-    rand_ = ThreadLocalRandom.current();
-  }
-  */
 
   // constructor used for deserialization and testing
   // does NOT copy the incoming ArrayList since this is an internal
--- a/src/main/java/org/apache/datasketches/sampling/EbppsItemsSample.java
+++ b/src/main/java/org/apache/datasketches/sampling/EbppsItemsSample.java
@@ -26,9 +26,6 @@ import java.util.ArrayList;
 import java.util.Random;
 import java.util.concurrent.ThreadLocalRandom;
 
-import org.apache.datasketches.common.ArrayOfItemsSerDe;
-import org.apache.datasketches.memory.Memory;
-
 // this is a supporting class used to hold the raw data sample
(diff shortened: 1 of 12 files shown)


src/main/java/org/apache/datasketches/sampling/EbppsItemsSketch.java
```diff
@@ -0,0 +1,539 @@
...
+   * probabilsitic, differing by at most 1 item.
+   * @return the current sketch sample
+   */
+  public ArrayList<T> getResult() { return sample_.getSample(); }
+
+  /**
+   * Provides a human-readable summary of the sketch
+   * @return a summary of information in the sketch
+   */
+  public String toString() {
```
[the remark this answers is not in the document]
jmalkin: Will fix upon next push

src/main/java/org/apache/datasketches/sampling/EbppsItemsSample.java
```diff
@@ -0,0 +1,301 @@
...
+  public ArrayList<T> getSample() {
+    final double cFrac = c_ % 1;
+    final boolean includePartial = partialItem_ != null && rand_.nextDouble() < cFrac;
+    final int resultSize = (data_ != null ? data_.size() : 0) + (includePartial ? 1 : 0);
+
+    if (resultSize == 0)
+      return null;
+
+    ArrayList<T> result = new ArrayList<>(resultSize);
+    for (T item : data_)
```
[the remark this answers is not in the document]
jmalkin: Fixed in next push

src/test/java/org/apache/datasketches/sampling/EbppsSketchTest.java
```diff
@@ -0,0 +1,308 @@
...
+    ArrayList<T> sample2 = sk2.getResult();
+
+    if (sk1.getC() < 1.0) {
+      if (sample1 != null && sample2 != null) {
+        assertEquals(sample1.size(), sample2.size());
+        assertEquals(sample1.get(0), sample2.get(0));
+      }
+      // nothing to test if one is null and the other isn't
+    } else {
+      final int len = Math.min(sample1.size(), sample2.size());
```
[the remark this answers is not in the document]
jmalkin: False positive since sk1.getC() == sk2.getC() and sk1.getC() >= 1.0 but added an extra assert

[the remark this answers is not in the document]
jmalkin: False positive since sk1.getC() == sk2.getC() and sk1.getC() >= 1.0 but added an extra assert

src/main/java/org/apache/datasketches/sampling/EbppsItemsSample.java
```diff
@@ -0,0 +1,301 @@
...
+        data_.add(item);  
+      }
+      partialItem_ = null;
+    } else {
+      data_ = null;
+      partialItem_ = item;
+    }
+  }
+
+  public void reset() {
```
leerho, line 85: You should be using Checkstyle. many of your public methods do not have javadocs, -or- they  don't need to be public.  Use the SketchesCheckstyle.xml in the /tools/ directory.

jmalkin: I can make everything in Sample package-private, but I think that's gonna make testing problematic when tests necessarily live in a different pacakge.

src/main/java/org/apache/datasketches/sampling/EbppsItemsSample.java
```diff
@@ -0,0 +1,301 @@
...
+    if (newC == newCInt)
+      partialItem_ = null;
+
+    c_ = newC;
+  }
+
+  public void merge(final EbppsItemsSample<T> other) {
+    //double cInt = Math.floor(c_);
+    double cFrac = c_ % 1;
+    double otherCFrac = other.c_ % 1;
```
leerho, line 188: Java modulo can produce negative numbers -- does that fact affect the following logic?

jmalkin: C is the expected sample size so it better not go negative!

src/main/java/org/apache/datasketches/sampling/EbppsItemsSketch.java
```diff
@@ -0,0 +1,540 @@
...
+      // data from sample_ -- itemBytes includes the partial item
+      mem.putDouble(EBPPS_C_DOUBLE, sample_.getC());
+      mem.putByteArray(EBPPS_ITEMS_START, itemBytes, 0, itemBytes.length);
+    }
+
+    return outArr;
+  }
+
+  private void checkK(final int k) {
+    if (k == 0 || k > MAX_K)
```
leerho, line 537: you may as well check if k is negative, i.e., 
if (k <= 0 || k > MAX_K) ...

src/main/java/org/apache/datasketches/sampling/EbppsItemsSketch.java
```diff
@@ -0,0 +1,540 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+import org.apache.datasketches.common.ArrayOfItemsSerDe;
+import org.apache.datasketches.common.Family;
+import org.apache.datasketches.common.SketchesArgumentException;
+import org.apache.datasketches.memory.Memory;
+import org.apache.datasketches.memory.WritableMemory;
+
+public class EbppsItemsSketch<T> {
```
leerho, line 35: Needs Javadocs

src/main/java/org/apache/datasketches/sampling/PreambleUtil.java
```diff
@@ -102,7 +102,7 @@
  *  </pre>
  *
  * <p><strong>VarOpt Union:</strong> VarOpt unions also store more information than a reservoir
- * sketch. As before, we keep values with similar o hte same meaning in corresponding locations
+ * sketch. As before, we keep values with similar to the same meaning in corresponding locations
```
leerho, line 105: You have multiple sketches using the same PreambleUtil class.  That is OK, but where you have the documentation for each of the specific sketches that use this class is confusing.  You chose to place the preamble structure for Ebpps in its Sketch class, but for VarOpt and Reservoir is in this class.  You could either move the specific preamble structures in to their respective sketch classes, or (I think I would prefer) have all of the structure docs in this class, separated by clear documentation boundaries.  I.e., use of <h2> </h2> headers, which work great in javadocs.

src/main/java/org/apache/datasketches/sampling/EbppsItemsSample.java
leerho: But from what I see, your tests are in the same package, so you could use package-private.

src/main/java/org/apache/datasketches/sampling/EbppsItemsSample.java
leerho: Is there a check somewhere that makes sure that C can never be negative?

leerho (commented):
- I have some questions about your fractional logic: somewhere we need to be sure that "c_" never is allowed to be negative.
- Also, are you using Checkstyle?
- A minor change to make sure k is not allowed to be negative.

src/main/java/org/apache/datasketches/sampling/EbppsItemsSample.java
leerho: Or are you referring to when we move to modules (java 17 +)?

src/main/java/org/apache/datasketches/sampling/EbppsItemsSample.java
jmalkin: Again, C must never go negative. I think the key is needing to check for non-negativity elsewhere which will prevent it from being negative here.

If someone modifies the value via reflection or something, yes, bad things will happen.

leerho: Yes.

src/main/java/org/apache/datasketches/sampling/EbppsItemsSketch.java
jmalkin: Oh! How did I forget that one! Yes, it definitely does.

src/main/java/org/apache/datasketches/sampling/PreambleUtil.java
jmalkin: I was unsure of the best way to do this. I feel like stacking them all together is not actually helpful since the class starts to feel bloated. You seem less worried about that?

leerho: If the sketch specific docs, and related code is nicely separated out with headers, it won't feel bloated.  It all has a reason for being there.   A class begins to feel "bloated", I think, when too much totally unrelated stuff is put into the same class.  

The intention was for the PreambleUtil class to be THE place to go to understand the various serialization formats and hold the static constants and static helper methods related to serialization. Having this stuff spread out in different classes makes it hard to find -- 'cause I, at least, will not be able to remember where to go to find it.  We have this inconsistency in some other sketches as well.

src/main/java/org/apache/datasketches/sampling/EbppsItemsSample.java
jmalkin: Just made everything package-private, but yes, it's going to be a problem with modules.

src/main/java/org/apache/datasketches/sampling/EbppsItemsSketch.java
jmalkin: Done

src/main/java/org/apache/datasketches/sampling/PreambleUtil.java
jmalkin: The challenge here is that I didn't want to create a stand-alone family, so there are now 3 separate sketches squashed into the same `PreambleUtil` class. But that's fine, I certainly don't have an objection to moving it.

jmalkin:
In theory I (now) have checkstyle running but it's not complaining if I create a public method w/o a comment. Not sure that VS Code is using it properly.

leerho (approved):
Thank you for your work on this!

PR #504, opened 2024-02-14
Minor changes to BoundsRule

leerho:


leerho:
This is already included in another branch that I want to merge instead.

PR #505, opened 2024-02-14
Improve partition boundaries

leerho:
This fixes a bug discovered in recent Druid testing.

Makes other improvements to the getPartionBoundaries code.

### DIFF
--- a/src/main/java/org/apache/datasketches/quantiles/ItemsSketchSortedView.java
+++ b/src/main/java/org/apache/datasketches/quantiles/ItemsSketchSortedView.java
@@ -92,7 +92,6 @@ public class ItemsSketchSortedView<T> implements GenericSortedView<T>, Partition
   ItemsSketchSortedView(final ItemsSketch<T> sketch) {
     if (sketch.isEmpty()) { throw new SketchesArgumentException(EMPTY_MSG); }
     this.totalN = sketch.getN();
-    final int k = sketch.getK();
     final int numQuantiles = sketch.getNumRetained();
     this.quantiles = (T[]) Array.newInstance(sketch.clazz, numQuantiles);
     this.minItem = sketch.minItem_;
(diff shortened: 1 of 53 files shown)


src/main/java/org/apache/datasketches/quantiles/ItemsSketchSortedView.java
```diff
@@ -155,27 +159,68 @@ public GenericPartitionBoundaries<T> getPartitionBoundaries(final int numEqually
...
+    final long[] partNatRanks = new long[partArrLen];
+    final double[] partNormRanks = new double[partArrLen];
+    //adjust ends
+    int adjLen = svLen;
+    final boolean adjLow = quantiles[0] != minItem;
+    final boolean adjHigh = quantiles[svLen - 1] != maxItem;
+    adjLen += adjLow ? 1 : 0;
+    adjLen += adjHigh ? 1 : 0;
+
+    final T[] adjQuantiles;
```
jmalkin, line 191: This is the start of a long code block with no indication of what it's doing. It'd be useful to have a little description of what's going on here.

src/test/java/org/apache/datasketches/sampling/EbppsSketchTest.java
```diff
@@ -22,8 +22,10 @@
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertFalse;
 import static org.testng.Assert.assertTrue;
+import static org.testng.Assert.fail;
 
 import java.util.ArrayList;
+import java.util.Arrays;
```
jmalkin, line 28: This is probably extraneous?

jmalkin (approved):
I don't know that I 100% get the tests in ItemsSketchPartitionBoundariesTest but I think they seemed fine.

Two minor comments (one a request for a comment, the other something that's probably not necessary). Nice to fix, not gonna block this on them.

src/test/java/org/apache/datasketches/sampling/EbppsSketchTest.java
leerho: Yes, and so is: import static org.testng.Assert.fail;
I will fix.

src/main/java/org/apache/datasketches/quantiles/ItemsSketchSortedView.java
leerho: Good feedback.  Will do.

PR #513, opened 2024-02-27
Bloom Filter

jmalkin:
A Bloom filter isn't quite one of our normal sketches, but we've had requests over the years for one. Why do we need yet another implementation of this? Comparing versus Spark's implementation (itself based somewhat on Guava's) I noticed a couple things:
1. The theoretical size can exceed 2^31-1 bits -- but only 31 bits of the hash function are ever used. The index is always a positive 32-bit int.
2. Our library specializes in simple cross-language portability. While it may be a good idea to look at alternatives at some point, seamless data movement between languages is a known quantity. When we port this to C++/Python, we'll have that in ways that are at least somewhat more complicated with other versions.

API change suggestions are quite welcome. I'm wondering if I should move the public constructors entirely to the builder class, for instance.

Since we have a couple newly donated membership filters, once we have API nomenclature down I plan to make a MembershipFilter abstract class or interface in the directory above so that we can have a common usage API. That should make using filters in distributed systems particularly useful -- deserialize any of them and blindly use regardless of the specific underlying implementation. That will make it easier for people to experiment with and, with luck, adopt the newer filters once they're production ready.


jmalkin:
FWIW, I tried allocating something with more than 2^31-1 bytes so requiring more than an int for the index) and it works fine from the command line. When I tested that in my IDE it complained about memory use, so I didn't make a test out of that.

jmalkin:
Wondering if I should move the package to something like `filters.bloom` rather than `filters.bloomfilter`. But if we do something like the infinifilter, i don't think we'd call if `filters.infini`.

Related, we could be a little more explicit and go with something like `membership-filter` but that feels a little long. Maybe clarity should win over brevity here.

src/main/java/org/apache/datasketches/filters/bloomfilter/BitArray.java
```diff
@@ -0,0 +1,219 @@
...
+    final int numLongs = (int) Math.ceil(numBits / 64.0);
+    numBitsSet_ = 0;
+    data_ = new long[numLongs];
+  }
+
+  // uses the provided array
+  BitArray(final long[] data) {
+    data_ = data;
+    numBitsSet_ = 0;
+    for (long val : data) {
```
AlexanderSaydakov, line 59: final?

src/main/java/org/apache/datasketches/filters/bloomfilter/BitArray.java
```diff
@@ -0,0 +1,219 @@
...
+      return new BitArray((long) numLongs * Long.SIZE);
+    }
+
+    final long[] data = new long[numLongs];
+    mem.getLongArray(DATA_OFFSET, data, 0, numLongs);
+    return new BitArray(data);
+  }
+
+  boolean isEmpty() {
+    return numBitsSet_ == 0;
```
AlexanderSaydakov, line 82: what if it is dirty?

src/main/java/org/apache/datasketches/filters/bloomfilter/BitArray.java
```diff
@@ -0,0 +1,219 @@
...
+
+  int getArrayLength() { return data_.length; }
+
+  // applies logical OR
+  void union(final BitArray other) {
+    if (data_.length != other.data_.length) {
+      throw new SketchesArgumentException("Cannot union bit arrays with unequal lengths");
+    }
+
+    long numBitsSet = 0;
```
AlexanderSaydakov, line 134: why not use the existing member variable? would it be slower?

src/main/java/org/apache/datasketches/filters/bloomfilter/BitArray.java
jmalkin: i fixed that in the branch

src/main/java/org/apache/datasketches/filters/bloomfilter/BitArray.java
jmalkin: in the branch I now accept numBitsSet as an input and don't bother recalculating it

src/main/java/org/apache/datasketches/filters/bloomfilter/BitArray.java
jmalkin: bias towards not modifying the member variable constantly, but i think that's silly of me

src/main/java/org/apache/datasketches/filters/bloomfilter/BloomFilter.java
```diff
@@ -0,0 +1,747 @@
...
+   *
+   * <p>Note: this will not produce the same output hash values as the {@link #update(char[])}
+   * method and will generally be a little slower depending on the complexity of the UTF8 encoding.
+   * </p>
+   *
+   * @param item an item with which to update the filter
+   */
+  public void update(final String item) {
+    if (item == null || item.isEmpty()) { return; }
+    final byte[] strBytes = item.getBytes(StandardCharsets.UTF_8);
```
AlexanderSaydakov, line 212: why not do this on toCharArray() skipping conversion?

jmalkin: What's going to give the most consistent cross-language behavior?

src/main/java/org/apache/datasketches/filters/bloomfilter/BloomFilter.java
```diff
@@ -0,0 +1,747 @@
...
+   * Serializes the current BloomFilter to an array of longs. Unlike {@link #toByteArray()},
+   * this method can handle any size filter.
+   *
+   * @return A serialized image of the current BloomFilter as long[]
+   */
+  public long[] toLongArray() {
+    final long sizeBytes = getSerializedSizeBytes();
+
+    final long[] longs = new long[(int) (sizeBytes >> 3)];
+    final WritableMemory wmem = WritableMemory.writableWrap(longs);
```
AlexanderSaydakov, line 709: there is Buffer (mem.asBuffer()) that keeps track of the offset, so we don't have to

src/main/java/org/apache/datasketches/filters/bloomfilter/BloomFilter.java
AlexanderSaydakov: right, for cross-language compatibility this should be better

src/main/java/org/apache/datasketches/filters/bloomfilter/BloomFilter.java
```diff
@@ -50,34 +49,24 @@
  * increase. The size of a Bloom filter will be linear in the expected number of
  * distinct items.</p>
  *
- * <p>See the BloomFilterBuilder class for methods to create a filter sized correctly
- * for a target number of distinct elements and a target false positive probability.</p>
+ * <p>See the BloomFilterBuilder class for methods to create a filter, especiall
```
AlexanderSaydakov, line 52: typo

AlexanderSaydakov (approved):

jmalkin:
I decided that adding a missing `y` in a comment was ok to include without a separate review.

PR #515, opened 2024-02-29
Changes based on review

jmalkin:
Including changes suggested from Slack chat. Separate branch so it doesn't reset the larger review state.


src/main/java/org/apache/datasketches/filters/bloomfilter/BitArray.java
```diff
@@ -103,7 +108,7 @@ boolean getAndSetBit(final long index) {
       return true; // already seen
     } else {
```
AlexanderSaydakov, line 109: unnecessary else

AlexanderSaydakov (approved):

PR #516, opened 2024-02-29
Improvements realized while analyzing Issue 514

leerho:
While analyzing Issue 514, I realized that adding a few new asserts and/or exceptions would be very helpful to users and to us when debugging.  Specifically, they consist of 3 different detections:
- Throwing an exception if a self-merge is detected. This is an easy-to-make user error and warrants an exception.

These next two should be very rare, but are also directly in the update path, so I think an _assert_ is more appropriate.
- Throwing an AssertionError if N (the number of items input into the sketch) exceeds Long.MAX_VALUE.
- Throwing an AssertionError if the number of levels exceed 60

Note that these new assertions and exception have only been implemented so far in KLL Doubles.  Implementation into the other sketches will be a later PR in order to keep the size of the PRs down.

I also realized that our built-in shuffle class could be improved by allowing generic arrays to be shuffled and also providing the option of a user-provided Random generator instead of only the internal default.

### DIFF
--- a/src/main/java/org/apache/datasketches/kll/KllDoublesHelper.java
+++ b/src/main/java/org/apache/datasketches/kll/KllDoublesHelper.java
@@ -123,8 +123,7 @@ final class KllDoublesHelper {
   }
 
   //assumes readOnly = false and UPDATABLE, called from KllDoublesSketch::merge
-  static void mergeDoubleImpl(final KllDoublesSketch mySketch,
-      final KllDoublesSketch otherDblSk) {
+  static void mergeDoubleImpl(final KllDoublesSketch mySketch, final KllDoublesSketch otherDblSk) {
     if (otherDblSk.isEmpty()) { return; }
 
     //capture my key mutable fields before doing any merging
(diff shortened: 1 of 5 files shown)


leerho:
This is a tough one. I discovered these issues while running malformed test code.  In all three cases, the sketch crashes.  
If the user just runs their environment with "-ea" enabled, the asserts will run.
Changing the asserts to exceptions means running a check always for a situation that should be exceedingly rare and is in the direct path of every insertion.

leerho:
This is a tough one. I discovered these issues running buggy test code. In the two assert cases the sketch will crash without any checks.  If the user reruns the environment with"-ea" the asserts will display the meaningful error message.
Converting to exceptions means the checks will always run for what should be extremely rare situations.   I hesitate to put too many checks directly in the update path.   There are many, many places in our code where we use longs and do not check for overflow, because longs are so huge.

leerho:
Removed two asserts and converted another one to throw ArithmeticException when the sum of two longs overflows.  Turns out that ` long Math.addExact(long,long) ` is the perfect solution. It tests for overflow and if true throws an ArithmeticException with the comment "long overflow", which is perfect.  And it is fast. I checked the Math source code.

This is a new exception, but it is thrown just before the sketch crashes.  So I do not believe this, on its own warrants a major release. However, once we are all done, we may have other reasons for a major release.

PR #521, opened 2024-03-07
Fixing SpotBugs exclusions in the /tools/FindBugsExcludeFilter.xml.

leerho:
Some of these were no longer needed because newer versions of SpotBugs have fixed earlier False Positive detections.

I was able to fix some just by getting a better understanding of what SpotBugs was complaining about.  Some of it was pretty subtle.

Nonetheless, this is a much shorter list of exclusions than what we had before.

### DIFF
--- a/src/main/java/org/apache/datasketches/theta/EmptyCompactSketch.java
+++ b/src/main/java/org/apache/datasketches/theta/EmptyCompactSketch.java
@@ -41,12 +41,12 @@ final class EmptyCompactSketch extends CompactSketch {
 
   private EmptyCompactSketch() {}
 
-  static EmptyCompactSketch getInstance() {
+  static synchronized EmptyCompactSketch getInstance() {
     return EMPTY_COMPACT_SKETCH;
   }
 
   //This should be a heapify
-  static EmptyCompactSketch getHeapInstance(final Memory srcMem) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/apache/datasketches/theta/EmptyCompactSketch.java
```diff
@@ -41,12 +41,12 @@ final class EmptyCompactSketch extends CompactSketch {
 
   private EmptyCompactSketch() {}
 
-  static EmptyCompactSketch getInstance() {
+  static synchronized EmptyCompactSketch getInstance() {
```
jmalkin, line 44: This is an issue with a private static final instance? Is the concern that it can be modified with reflection?

jmalkin (approved):
I do question the need for `synchronized` when the only thing the method does is return a private static final object...

src/main/java/org/apache/datasketches/theta/EmptyCompactSketch.java
leerho: Yes it is possible to circumvent using reflection.  There are more complex solutions which prevent this, but, given that we have never claimed that our library is thread-safe, or immune to reflection I think this simple modification at least satisfies the SpotBugs detected issue so that we are not completely ignoring it.

PR #531, opened 2024-03-19
no weights in buffer, simplified sizing

AlexanderSaydakov:


jmalkin (approved):
Similar to related c++ changes, although in this case removing bufferedWeight_

PR #533, opened 2024-03-21
Allow updating sketches with ByteBuffers

### LINKED ISSUE #510: Allow UpdateSketch#update to take a byte array slice
UpdateSketch#update(byte[]) passes its byte array argument to MurmurHash3#hash(byte[],int,int,long) but always uses an offset of 0 and a length of data.length.  Adding a new update method that also takes an offset and a length would allow users to update sketches without having to copy data if their inputs are already references into large byte arrays.

For context, I'm coming from Apache Lucene where term data is almost always passed around as references into these types of large byte arrays, and being able to update sketches like this would be very helpful in cutting down memory copies.  If the maintainers think this would be useful to others I'm happy to open a pull request.

romseygeek:
Fixes #510


leerho (approved):
Nice Job!

PR #535, opened 2024-03-26
This extends the usability of the Kolmogorov-Smirnov test

leerho:
from the Classic DoublesSketch to KllDoublesSketch and the KllFloatsSketch.

In order to do this I needed to add "getNormalizedRankError(boolean) to the root interface QuantilesAPI. All of the sketches (except REQ) had this anyway. In REQ I just throw an unsupported operation exception, because it makes no sense for REQ.

This also means the KolmogorovSmirnov class and its test move to the "quantilescommon" package.


src/test/java/org/apache/datasketches/quantilescommon/KolmogorovSmirnovTest.java
```diff
@@ -0,0 +1,355 @@
...
+
+  @Test
+  public void printlnTest() {
+    println("PRINTING: "+this.getClass().getName());
+  }
+
+  /**
+   * @param s value to print
+   */
+  static void println(String s) {
```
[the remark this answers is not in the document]
leerho: Fixed.

jmalkin:
I don't think it's that it makes no sense for REQ, but it won't be able to share code nicely. With classic quantiles and KLL we can share an API since error is constant over the sketch. For REQ we'd need a point-specific error estimate.

Completely out of scope of this PR though.

src/main/java/org/apache/datasketches/req/ReqSketch.java
```diff
@@ -235,6 +235,15 @@ public long getN() {
     return totalN;
   }
 
+  @Override
```
jmalkin, line 238: Didn't realize `@Override` could appear above a comment like this. Makes sense in that comments are stripped when compiling.

src/main/java/org/apache/datasketches/quantilescommon/KolmogorovSmirnov.java
```diff
@@ -0,0 +1,167 @@
...
+   * @param sketch1 first Input QuantilesAPI
+   * @param sketch2 second Input QuantilesAPI
+   * @param tgtPvalue Target p-value. Typically .001 to .1, e.g., .05.
+   * @return Boolean indicating whether we can reject the null hypothesis (that the sketches
+   * reflect the same underlying distribution) using the provided tgtPValue.
+   */
+  public static boolean kolmogorovSmirnovTest(final QuantilesAPI sketch1,
+      final QuantilesAPI sketch2, final double tgtPvalue) {
+
+    final double delta = isDoubleType(sketch1, sketch2)
```
jmalkin, line 150: Kind of unfortunate we can't mix types :(

src/test/java/org/apache/datasketches/quantilescommon/KolmogorovSmirnovTest.java
```diff
@@ -0,0 +1,357 @@
...
+
+   final int n =  (3 * k) - 1;
+   for (int i = 0; i < n; ++i) {
+     final double x = rand.nextGaussian();
+     s1.update(x + 500);
+     s2.update(x);
+   }
+   final double delta = DoublesSketch.getNormalizedRankError(k, false);
+   println("D = " + KolmogorovSmirnov.computeKSDelta(s1, s2));
+   assertEquals(KolmogorovSmirnov.computeKSDelta(s1, s2), 1.0, delta);
```
jmalkin, line 53: `delta` is not the correct test target here. For disjoint distributions it needs to be much, much smaller than that. Maybe not quite 0.0 but very close.

src/test/java/org/apache/datasketches/quantilescommon/KolmogorovSmirnovTest.java
```diff
@@ -0,0 +1,357 @@
...
+
+   final int n =  (3 * k) - 1;
+   for (int i = 0; i < n; ++i) {
+     final double x = rand.nextGaussian();
+     s1.update(x + 500);
+     s2.update(x);
+   }
+   final double delta = KllSketch.getNormalizedRankError(k, false);
+   println("D = " + KolmogorovSmirnov.computeKSDelta(s1, s2));
+   assertEquals(KolmogorovSmirnov.computeKSDelta(s1, s2), 1.0, delta);
```
jmalkin, line 72: Same as above. `delta` is much too large to test anything meaningful here.

src/test/java/org/apache/datasketches/quantilescommon/KolmogorovSmirnovTest.java
```diff
@@ -0,0 +1,357 @@
...
+
+   final int n =  (3 * k) - 1;
+   for (int i = 0; i < n; ++i) {
+     final float x = (float)rand.nextGaussian();
+     s1.update(x + 500);
+     s2.update(x);
+   }
+   final double delta = KllSketch.getNormalizedRankError(k, false);
+   println("D = " + KolmogorovSmirnov.computeKSDelta(s1, s2));
+   assertEquals(KolmogorovSmirnov.computeKSDelta(s1, s2), 1.0, delta);
```
jmalkin, line 91: Again

src/test/java/org/apache/datasketches/quantilescommon/KolmogorovSmirnovTest.java
```diff
@@ -0,0 +1,357 @@
...
+
+   final int n =  (3 * k) - 1;
+   for (int i = 0; i < n; ++i) {
+     final double x = rand.nextGaussian();
+     s1.update(x);
+     s2.update(x);
+   }
+   final double delta = DoublesSketch.getNormalizedRankError(k, false);
+   println("D = " + KolmogorovSmirnov.computeKSDelta(s1, s2));
+   assertEquals(KolmogorovSmirnov.computeKSDelta(s1, s2), 0, delta);
```
jmalkin, line 158: Here, `delta` is too restrictive in general. It may be fine with the given seed, but conceptually it's the wrong thing to test against. The allowable margin is larger than the single-sketch delta.

Reading test code is a good way to learn how things are supposed to work, so we shouldn't have our tests suggest that this is a good target to check against. Or at the very least a comment noting that we're using it despite it being too tight a bound for this test in general.

src/test/java/org/apache/datasketches/quantilescommon/KolmogorovSmirnovTest.java
```diff
@@ -0,0 +1,357 @@
...
+
+   final int n =  (3 * k) - 1;
+   for (int i = 0; i < n; ++i) {
+     final double x = rand.nextGaussian();
+     s1.update(x);
+     s2.update(x);
+   }
+   final double delta = KllSketch.getNormalizedRankError(k, false);
+   println("D = " + KolmogorovSmirnov.computeKSDelta(s1, s2));
+   assertEquals(KolmogorovSmirnov.computeKSDelta(s1, s2), 0, delta);
```
jmalkin, line 177: same

src/test/java/org/apache/datasketches/quantilescommon/KolmogorovSmirnovTest.java
```diff
@@ -0,0 +1,357 @@
...
+
+   final int n =  (3 * k) - 1;
+   for (int i = 0; i < n; ++i) {
+     final float x = (float)rand.nextGaussian();
+     s1.update(x);
+     s2.update(x);
+   }
+   final double delta = KllSketch.getNormalizedRankError(k, false);
+   println("D = " + KolmogorovSmirnov.computeKSDelta(s1, s2));
+   assertEquals(KolmogorovSmirnov.computeKSDelta(s1, s2), 0, delta);
```
jmalkin, line 196: same

jmalkin (changes requested):

src/main/java/org/apache/datasketches/req/ReqSketch.java
leerho: no comment

src/main/java/org/apache/datasketches/quantilescommon/KolmogorovSmirnov.java
leerho: In C++, yes, in Java, no.

src/test/java/org/apache/datasketches/quantilescommon/KolmogorovSmirnovTest.java
leerho: and you made Random deterministic.  I overlooked that.

jmalkin: Looking at this further, it's a pre-existing issue with our KS test. We can look at how to improve that, but it's not an issue with extending the test to KLL.

leerho: After a long side conversation, resolved to change name delta -> eps and the tolerance to 2 * eps.

src/test/java/org/apache/datasketches/quantilescommon/KolmogorovSmirnovTest.java
leerho: After a long side conversation, resolved to change name delta -> eps and the tolerance to 2 * eps.

src/test/java/org/apache/datasketches/quantilescommon/KolmogorovSmirnovTest.java
leerho: After a long side conversation, resolved to change name delta -> eps and the tolerance to 2 * eps.

src/test/java/org/apache/datasketches/quantilescommon/KolmogorovSmirnovTest.java
leerho: After a long side conversation, resolved to change name delta -> eps and the tolerance to 2 * eps.

jmalkin (approved):
We need to adjust things to get the tolerance for the disjoint tests way down, but that's a separate issue. This looks good.

PR #538, opened 2024-03-30
Direct implementation of Bloom filter

jmalkin:
Adds read-only and writable direct implementations of the Bloom filter.

Especially if reading very large filters just to query them, the ability to wrap the filter cuts the required memory footprint in half.

### DIFF
--- a/src/main/java/org/apache/datasketches/filters/bloomfilter/BitArray.java
+++ b/src/main/java/org/apache/datasketches/filters/bloomfilter/BitArray.java
@@ -19,7 +19,10 @@
 
  package org.apache.datasketches.filters.bloomfilter;
 
- /**
+import org.apache.datasketches.memory.Memory;
+import org.apache.datasketches.memory.WritableMemory;
+
+/**
   * This class holds an array of bits suitable for use in a Bloom Filter
   *
   * <p>Rounds the number of bits up to the smallest multiple of 64 (one long)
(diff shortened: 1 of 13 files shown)


src/main/java/org/apache/datasketches/filters/bloomfilter/BitArray.java
```diff
@@ -17,194 +17,111 @@
...
   // prints the raw BitArray as 0s and 1s, one long per row
   @Override
   public String toString() {
     final StringBuilder sb = new StringBuilder();
-    for (int i = 0; i < data_.length; ++i) {
+    for (int i = 0; i < getArrayLength(); ++i) {
       sb.append(i + ": ")
-        .append(printLong(data_[i]))
+        .append(printLong(getLong(i)))
         .append("\n");
```
leerho, line 88: Just as a comment: we should always use the Java "line.separator" instead of "\n" for new lines, because the Java one is automatically adjusted based on the platform.  "\n" works on unix and MacOS, but Windows requires "\r\n".
I am not perfect in this either, but try to keep it in mind. You can acquire the Java "line.separator" like this:
`public static final String LS = System.getProperty("line.separator");`

leerho (approved):
Looks good.

src/main/java/org/apache/datasketches/filters/bloomfilter/BitArray.java
jmalkin: Thanks for catching this. I forgot to update `toString()` properly after my initial uses for debugging in other ways, too, but missed this one on a subsequent pass, too.

PR #544, opened 2024-04-08
Fix raw line feeds in main.

leerho:


jmalkin (approved):
i missed a couple `final`s previously. thanks for catching those!

PR #545, opened 2024-04-12
Fix memory leak.

leerho:
The WritableMemory.allocateDirect returns a WritableHandle that implements AutoCloseable.  And just as in C, when you malloc memory you are responsible to release that memory.  In this case it is done with a _try-with-resources_ block. As you will see in the fix.


jmalkin (approved):
Oops. Thank you. Was mostly just trying to make sure I used allocateDirect somewhere to be sure there wasn't weirdness where it'd work for on-heap but not off-heap memory.

PR #549, opened 2024-04-17
Ks updates

leerho:
In starting the release process I found this dangling commit that I forgotten about and never merged.  It just simplifies the readability of the KolmogorovSmirnov.java function by importing the all the static Math functions, which makes the code a little easier to read.


jmalkin (approved):
Ship it!

PR #550, opened 2024-04-17
Propagation of change in ArrayOfItemsSerDe.

leerho:
the last commit requires a propagation to the other ArrayOf-type-SerDe classes


jmalkin (approved):
rare that we've had a PR with only deletions!

PR #551, opened 2024-04-18
The tiny bloomfilter change is to fix a SpotBugs error.

leerho:
The POM change updates maven and most of the plugins to their latest versions.


jmalkin (approved):
Not gonna pretend I can say whether the version changes are necessary/correct but they are at least all newer versions.

PR #554, opened 2024-04-28
Add rough memory size tracking with KllItemsSketch

ZacBlanco:
When generating KllSketches, systems may need to have an idea about how much memory utilization there is for a particular sketch. For sketches with fixed-width types the answer can be computed efficiently.

With the KllItemsSketch, this is more difficult because the sketch can support String-types with variable widths. This commit adds implementation support to expose the `getTotalItemsNumBytes` method so that external systems can roughly track the memory utilization of a particular sketch.

The change accomplishes this by intercepting the code where a new item is added to the items array, or when a new array is generated entirely. This will add a slight overhead due to the sketch now needing to compute the length of inputs. For fixed-width types the overhead is low. For string this will require a call to encode the string as UTF-8/16 before adding it to the array.

For fixed-width types, the calculations have little effective overhead as the computation is a single array-access lookup + multiplication with the type width.

### DIFF
--- a/src/main/java/org/apache/datasketches/common/ArrayOfBooleansSerDe.java
+++ b/src/main/java/org/apache/datasketches/common/ArrayOfBooleansSerDe.java
@@ -122,4 +122,10 @@ public class ArrayOfBooleansSerDe extends ArrayOfItemsSerDe<Boolean> {
 
   @Override
   public Class<Boolean> getClassOfT() { return Boolean.class; }
+
+  @Override
+  public boolean isFixedWidth()
+  {
+    return true;
+  }
 }
--- a/src/main/java/org/apache/datasketches/common/ArrayOfDoublesSerDe.java
+++ b/src/main/java/org/apache/datasketches/common/ArrayOfDoublesSerDe.java
@@ -101,4 +101,10 @@ public class ArrayOfDoublesSerDe extends ArrayOfItemsSerDe<Double> {
 
   @Override
   public Class<Double> getClassOfT() { return Double.class; }
+
+  @Override
+  public boolean isFixedWidth()
+  {
+    return true;
+  }
 }
--- a/src/main/java/org/apache/datasketches/common/ArrayOfItemsSerDe.java
+++ b/src/main/java/org/apache/datasketches/common/ArrayOfItemsSerDe.java
@@ -114,4 +114,9 @@ public abstract class ArrayOfItemsSerDe<T> {
    * @return the concrete class of type T
    */
   public abstract Class<T> getClassOfT();
+
+  /**
+   * @return if this class serializes all types to a fixed width.
+   */
+  public abstract boolean isFixedWidth();
 }
--- a/src/main/java/org/apache/datasketches/common/ArrayOfLongsSerDe.java
+++ b/src/main/java/org/apache/datasketches/common/ArrayOfLongsSerDe.java
@@ -100,4 +100,10 @@ public class ArrayOfLongsSerDe extends ArrayOfItemsSerDe<Long> {
 
   @Override
   public Class<Long> getClassOfT() { return Long.class; }
+
+  @Override
+  public boolean isFixedWidth()
+  {
+    return true;
+  }
 }
--- a/src/main/java/org/apache/datasketches/common/ArrayOfNumbersSerDe.java
+++ b/src/main/java/org/apache/datasketches/common/ArrayOfNumbersSerDe.java
@@ -240,4 +240,10 @@ public class ArrayOfNumbersSerDe extends ArrayOfItemsSerDe<Number> {
 
   @Override
   public Class<Number> getClassOfT() { return Number.class; }
+
+  @Override
+  public boolean isFixedWidth()
+  {
+    return false;
+  }
 }
--- a/src/main/java/org/apache/datasketches/common/ArrayOfStringsSerDe.java
+++ b/src/main/java/org/apache/datasketches/common/ArrayOfStringsSerDe.java
@@ -130,4 +130,10 @@ public class ArrayOfStringsSerDe extends ArrayOfItemsSerDe<String> {
 
   @Override
   public Class<String> getClassOfT() { return String.class; }
+
+  @Override
+  public boolean isFixedWidth()
+  {
+    return false;
+  }
 }
--- a/src/main/java/org/apache/datasketches/common/ArrayOfUtf16StringsSerDe.java
+++ b/src/main/java/org/apache/datasketches/common/ArrayOfUtf16StringsSerDe.java
@@ -124,4 +124,10 @@ public class ArrayOfUtf16StringsSerDe extends ArrayOfItemsSerDe<String> {
 
   @Override
   public Class<String> getClassOfT() { return String.class; }
+
+  @Override
+  public boolean isFixedWidth()
+  {
+    return false;
+  }
 }
--- a/src/main/java/org/apache/datasketches/kll/KllDirectCompactItemsSketch.java
+++ b/src/main/java/org/apache/datasketches/kll/KllDirectCompactItemsSketch.java
@@ -234,6 +234,12 @@ final class KllDirectCompactItemsSketch<T> extends KllItemsSketch<T> {
   }
 
   @Override
+  int getTotalItemsNumBytesInternal()
+  {
+    return getRetainedItemsSizeBytes();
+  }
+
+  @Override
   WritableMemory getWritableMemory() {
     return (WritableMemory)mem;
   }
--- a/src/main/java/org/apache/datasketches/kll/KllDoublesSketch.java
+++ b/src/main/java/org/apache/datasketches/kll/KllDoublesSketch.java
@@ -514,7 +514,7 @@ public abstract class KllDoublesSketch extends KllSketch implements QuantilesDou
   abstract byte[] getTotalItemsByteArr();
 
   @Override
-  int getTotalItemsNumBytes() {
+  public int getTotalItemsNumBytes() {
     return levelsArr[getNumLevels()] * Double.BYTES;
   }
 
--- a/src/main/java/org/apache/datasketches/kll/KllFloatsSketch.java
+++ b/src/main/java/org/apache/datasketches/kll/KllFloatsSketch.java
@@ -514,7 +514,7 @@ public abstract class KllFloatsSketch extends KllSketch implements QuantilesFloa
   abstract byte[] getTotalItemsByteArr();
 
   @Override
-  int getTotalItemsNumBytes() {
+  public int getTotalItemsNumBytes() {
     return levelsArr[getNumLevels()] * Float.BYTES;
   }
 
--- a/src/main/java/org/apache/datasketches/kll/KllHeapItemsSketch.java
+++ b/src/main/java/org/apache/datasketches/kll/KllHeapItemsSketch.java
@@ -53,6 +53,10 @@ final class KllHeapItemsSketch<T> extends KllItemsSketch<T> {
   private T minItem;
   private T maxItem;
   private Object[] itemsArr;
+  /**
+   * Roughly tracks the total stored size of items in-memory within the sketch.
+   */
+  private int totalItemsSize = 0;
 
   /**
    * New instance heap constructor.
@@ -131,6 +135,7 @@ final class KllHeapItemsSketch<T> extends KllItemsSketch<T> {
       this.minItem = item;
       this.maxItem = item;
       itemsArr[k - 1] = item;
+      totalItemsSize = serDe.sizeOf(item);
     } else if (memStruct == COMPACT_FULL) {
       int offset = DATA_START_ADR + memVal.numLevels * Integer.BYTES;
       this.minItem = serDe.deserializeFromMemory(srcMem, offset, 1)[0];
@@ -140,6 +145,7 @@ final class KllHeapItemsSketch<T> extends KllItemsSketch<T> {
       final int numRetained = levelsArr[memVal.numLevels] - levelsArr[0];
       final Object[] retItems = serDe.deserializeFromMemory(srcMem, offset, numRetained);
       System.arraycopy(retItems, 0, itemsArr, levelsArr[0], numRetained);
+      totalItemsSize = serDe.sizeOf((T[]) retItems);
     } else { //memStruct == UPDATABLE
       throw new SketchesArgumentException(UNSUPPORTED_MSG + "UPDATABLE");
     }
@@ -261,6 +267,12 @@ final class KllHeapItemsSketch<T> extends KllItemsSketch<T> {
   }
 
   @Override
+  int getTotalItemsNumBytesInternal()
+  {
+    return totalItemsSize;
+  }
+
+  @Override
   WritableMemory getWritableMemory() {
     return null;
   }
@@ -290,12 +302,26 @@ final class KllHeapItemsSketch<T> extends KllItemsSketch<T> {
 
   @Override
   void setItemsArray(final Object[] itemsArr) {
+    int sumItemSize = 0;
+    for (Object o : itemsArr) {
+      if (o == null) {
+        continue;
+      }
+      sumItemSize += serDe.sizeOf((T) o);
+    }
+    this.totalItemsSize = sumItemSize;
     this.itemsArr = itemsArr;
   }
 
   @Override
   void setItemsArrayAt(final int index, final Object item) {
+    if (itemsArr[index] != null) {
+      this.totalItemsSize -= serDe.sizeOf((T) itemsArr[index]);
+    }
     this.itemsArr[index] = item;
+    if (item != null) {
+      this.totalItemsSize += serDe.sizeOf((T) itemsArr[index]);
+    }
   }
 
   @Override
--- a/src/main/java/org/apache/datasketches/kll/KllItemsSketch.java
+++ b/src/main/java/org/apache/datasketches/kll/KllItemsSketch.java
@@ -353,8 +353,6 @@ public abstract class KllItemsSketch<T> extends KllSketch implements QuantilesGe
   @Override
   abstract int getRetainedItemsSizeBytes();
 
-  //abstract Object[] getRetainedItemsArray();
-
   @Override
   ArrayOfItemsSerDe<T> getSerDe() { return serDe; }
 
@@ -378,9 +376,16 @@ public abstract class KllItemsSketch<T> extends KllSketch implements QuantilesGe
   }
 
   @Override
-  int getTotalItemsNumBytes() {
-    throw new SketchesArgumentException(UNSUPPORTED_MSG);
+  public int getTotalItemsNumBytes() {
+    // if empty, exception is thrown.
+    if (serDe.isFixedWidth() && !isEmpty()) {
+      return levelsArr[getNumLevels()] * serDe.sizeOf(getMinItem());
+    } else {
+      return getTotalItemsNumBytesInternal();
+    }
   }
+
+  abstract int getTotalItemsNumBytesInternal();
 
   @Override
   void incNumLevels() {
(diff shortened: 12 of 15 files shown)


jmalkin:
Can you elaborate more on what problem this is trying to solve? Why would `getSerializedSizeBytes()` not work, for instance, which would have zero speed impact on updates, especially for non-generics? And that would provide a more accurate representation of the space used. This approach slows down all instances and is guaranteed to under-estimate the memory use.

ZacBlanco:
Hi Jon,

Thanks for your quick response. I think this change presents little overhead except for the ItemsSketch which uses non-fixed fixed-width types - specifically on Strings. I’m willing to iterate on this to find a solution that the community would be willing to accept. I don’t expect the PR to be accepted in the current state.

First, I’ll try to paint a picture of why I think this change is needed. Next, I’ll also provide a few of the microbenchmarks comparison between the master branch and this PR for a number of operations to show the impact of the PR on the performance.

----

### Background

The system we are trying to use KLL sketches in is large distributed database engine. Tables can easily have 100TB+ of data. We’re implementing a KLL Sketch aggregation function for a variety different data types. Fixed-width types work great, but we’re also trying to implement them for strings.

Imagine a table with 1T rows, then you run a `GROUP BY` query on that table. The cardinality of the GROUP BY is somewhere in the range of 100M.  The execution engine splits up the tasks and tries to schedule those tasks in a way that satisfies the memory constraints of the cluster. If you have 100M KLL sketches, depending on the parameters, it could amount to a not-insignificant chunk of memory and may need to make a task scheduling decision to move execution to another node based on that information. Hence, the system needs to track roughly how much memory an aggregation is using.

The design of the database engine is such that querying the current estimated size needs to be very fast, which means we need to incrementally track the size of KLL sketches as items are added. In the JVM, exact memory tracking is very expensive, but even a good ballpark number is useful for us scheduling these tasks.

In the current iteration of the sketch function implementation, we tried something like this:

```java
memoryUsage -= sketch.getSerializedSizeBytes();
sketch.update(item);
memoryUsage += sketch.getSerializedSizeBytes();
```

Unfortunately, `getSerializedSizeBytes()` is so expensive, it slows the queries down to a crawl for all types in the ItemsSketch. It’s over ~100X slower to do this than to simply call `sketch.update`. One of the main culprits which causes the slowdown is that for ItemsSketch, we end up doing a potentially large array copy for every call. The call chain is:


https://github.com/apache/datasketches-java/blob/0bff44b13ccf2eb27d2519942016d9d83327af5f/src/main/java/org/apache/datasketches/kll/KllSketch.java#L320

which calls:

https://github.com/apache/datasketches-java/blob/0bff44b13ccf2eb27d2519942016d9d83327af5f/src/main/java/org/apache/datasketches/kll/KllHeapItemsSketch.java#L235

which calls:

https://github.com/apache/datasketches-java/blob/0bff44b13ccf2eb27d2519942016d9d83327af5f/src/main/java/org/apache/datasketches/kll/KllHeapItemsSketch.java#L235


`getRetainedItemsByteArr` is doing a very large and useless allocation which makes getting the current size far too expensive. 

I hope the above explanation clears up the motivation behind this PR. Please feel free to ask any questions.

### Performance

I wrote some basic microbenchmarks with JMH to test the performance of these methods. The patch with benchmarks can be found here: https://github.com/ZacBlanco/datasketches-java/commit/102e383d72edd044e161c96e3031e28a303b14f5. The benchmarks test the performance of doubles, floats, items (strings and doubles) for the methods: update, getSerializedSizeBytes, and getTotalItemsNumBytes. The results are below.

image
image

You can see that aside from the update for `KllItemsSketch<String>`, the performance of virtually every other method is unchanged. Additionally, the cost of getTotalItemsNumBytes is multiple orders of magnitude lower than the cost of `getSerializedSizeBytes()`(Doubles: ~615 vs 10ns, Strings, ~200k ns vs 9ns), which solves my initial problem of having a quick way to get the rough memory usage of the ItemsSketch.

### Additional Optimizations

I think there are two related, but slightly different optimizations that can be made. 

First, in the ArrayOfStringsSerDe, to calculate the size of the byte array, the `sizeOf` method utilizes `String#getBytes`, which copies the existing String into a new byte array, and then returns the new array where only the length of the array is queried. This is quite slow and results in many more allocations than necessary. I would highly recommend the library adopt a method which calculates the length of the strings without these additional allocations.

There is an implementation in the Guava library which can probably be directly copied and dropped in to the datasketches library without having to pull in any dependencies. https://github.com/google/guava/blob/9596c996098272f0ea69cfdf45c7c1fe1ef8406d/guava/src/com/google/common/base/Utf8.java#L49

While I think there is no downside with this improvement, it doesn’t have a large enough performance impact to make `getSerializedSizeBytes` fast enough for me to use.

Second, I tested with tracking the size of `getTotalNumBytes` simply using `String#length` rather than doing a UTF-8 encoding inside of ArrayOfStringSerde#sizeOf. With that change, the update performance for the Strings sketch was virtually no different than without the size tracking. Unfortunately though, it means that we wouldn’t be able to support UTF-8 strings. Do you think there would be a good way to introduce using the String#length call to perform the incremental size tracking?

Thanks for taking the time to read. Please let me know your thoughts.

leerho:
Hi Zac, 
Thank you very much for your suggestion! It really helps us understand the issues you face in using our sketches and how we can improve the library!

The first concern that jumps out at me is that your specific problem and the solution you propose is focused on items as Strings. Although I recognize that Strings may be a dominant use case for the items sketches, it is not the only application, as items sketches can be used for arbitrary objects.  So I am hesitant to put into the sketch a lot of special code that is only useful for Strings. 

That being said, let me ask you some questions and explore some other ideas with you. In your write-up above you said:

> ...the system needs to track roughly how much memory an aggregation is using.

- This gives me hope!  Nonetheless, how "rough" is this need?  +/- 1%, 10%, 50% ? Because I think we can come up with a good approximation of the memory usage that may be much simpler than wedging a bunch of code for a specific use case into the sketch itself.

- You gave some numbers above.  Allow me to noodle with them (although I'm sure you did not want me to do this! )
    - Total data: 100TB = 1e14 bytes
    - Rows: 1T = 1e12 => ~ 100 bytes / row
    - String items per row ? I assume 1 ?
    - Groups: 100M = 1e8 groups => ~ 1e4 rows / group => 1e6 bytes / group

- Do these extrapolations sound about right?  If you are not sure, would it be possible for you to characterize your string length, row size, and group size distributions of your raw data?  This would be just three KllFloatsSketches, one each for capturing the string lengths, the row sizes and the group sizes.   Easier said than done, I'm sure, but this will be very useful, as I'll explain.

But first:

- What value of K are you using for your sketches now? And what is your requirement for rank accuracy that led you to choose this value of K?
- What are you using the sketches for?  
    - Data partitioning perhaps?  If so, we have a whole lot more to discuss!
    - Anomaly detection?
    - How does knowing the data distribution per group (per day, per whatever?) help you?
- I assume, although you didn't say exactly, that having an estimate of how much memory your aggregation functions are using is important for memory planning and partitioning, and more important than the fact that they are using memory. Is that correct?

--IDEA--
Sketches are statistical animals, so lets try to take advantage of statistics of your data to get approximate, yet useful estimations of the data you need for your memory planning.

The KLL sketches have a very well defined growth plan that looks something like the following. The red curve is for KllDoublesSketch and the blue curve is for KllFloatsSketch.

GrowthFloats Doubles_K200

Normalizing these two curves by the size of a double (8 bytes) and a float (4 bytes) produces this curve:

NormalizedGrowth_K200b

The message from these graphs is that once KLL sketches reach a certain size they grow very, very slowly. As you can see, from the red KllDoublesSketch curve, from 1M input doubles to 1B input doubles, the sketch only grows from 5KB to about 5.7KB or about 700 bytes.  And going from 1B input doubles to 1T input doubles will cause the size to increase by  another 700 bytes!  This is because that final slope remains constant.

Now with the second normalized curve above, and a distribution of item sizes extracted from your actual data, we can create a statistical model that pretty closely estimates the actual memory usage of a KllItemsSketch being fed the same data.  This approach could be used no matter what the item is, and it can be done while the KllItemsSketch is being feed data.  This might mean having two sketches side by side: A floats sketch tracking the distribution of item sizes, and the actual items sketch tracking the distribution of the items themselves.

Would you be willing to try something like this?

Cheers,
Lee.

leerho:
Zac, Thank you!  This is very helpful.  

I am not discarding your suggested improvements to the ArrayOfStringsSerDe.  The first one is especially important and we will definitely integrate that in the next release.  

The second one is more of a demonstration of where the problem is. Of course you are free to extend these ArrayOf*SerDe classes to suit. Substituting the char[].length for the UTF-8 byte[].length is only useful in English speaking countries where ASCII predominates.  Nonetheless, there are lots of situations where this assumption will not work.  

As to your suggestions in the PR, I will have to study them more closely.  Some I think are OK, others I will have to think about.

****
Back to your description of your use-case.  Allow me to try to play back to you what I am gathering from your description:

- Start with a very large table with ~1T (1e12) rows, each row can have several hundred columns.
- "With a string-size of 64 bytes and 1T rows, you're looking at 64T of data. Across 100M groups that's roughly 640k of data per group and 100K items per sketch in the group."
    - Doesn't this assume you are analyzing only one column?   
- Nonetheless, this data per group will vary widely and likely be power-law distributed. (a few groups with millions of rows, and millions of groups with only a few rows).

It is my understanding that it is at this group level where you want to analyze the distribution of items, where you might have a KLL sketch per column, thus many sketches just for this group.  And, of course, all the 100M groups would be configured with sketches as well implying a huge number of sketches for the entire query analysis  -- thus the concern about memory usage.

Assuming that the above is roughly right, here is where I am a little confused (I am not a DB engine expert by any means!)  Above you mentioned:

> The main objective of using these sketches is to estimate the result of predicates=, !=, <, <=, >, >= etc applied to columns on a table which enables the optimizer's CBO to create better query plans.

- Can you give me an example of how you use the distribution information from these sketches along with the user chosen predicates to help you with query optimization? 
- What are you looking for in the distributions from these sketches?
- What decisions can you (or your query planning model ) make from these distributions?



_Note: If you are looking for Heavy-Hitters the Frequent Items sketch might be very useful._

_Note: If you are converting these sketches into PMF histograms, I should caution you about trying to used too many split-points for the given K of the sketch, because you can easily exceed the error threshold of the sketch (producing garbage).  Our most recent release 6.0.0 has some extra functions to help you with this._  

*****
Given your estimate of an average of 64B per item, this is what the KLL Growth Path might look like:

NormalizedGrowth_64B_K200

This is obviously much bigger than 5KB!  This is just the previous graph, which was normalized to one byte, times 64.  

Cheers,
Lee.

ZacBlanco:
> With a string-size of 64 bytes and 1T rows, you're looking at 64T of data. Across 100M groups that's roughly 640k of data per group and 100K items per sketch in the group."
>    Doesn't this assume you are analyzing only one column?

Yes, that example was just for one column, but it depends on the user query as to whether or not we're creating sketches for one column, or multiple. I was just using one column to demonstrate a simple scenario for memory usage.

> It is my understanding that it is at this group level where you want to analyze the distribution of items, where you might have a KLL sketch per column, thus many sketches just for this group. And, of course, all the 100M groups would be configured with sketches as well implying a huge number of sketches for the entire query analysis -- thus the concern about memory usage.

Yes, this is correct. You could imagine this situation arises when a user writes something like:

```sql
SELECT kll_sketch(<column_x>), kll_sketch(<column_y>), ... FROM <table> GROUP BY <column1>, <column2>, ...
```

and `<table>` is the hypothetically very large table. I couldn't tell you exactly why a user might write a query like this, but it's something we have to support. We don't want to engine to crash if something like this were to be submitted.

> Can you give me an example of how you use the distribution information from these sketches along with the user chosen predicates to help you with query optimization?
>    What are you looking for in the distributions from these sketches?
>    What decisions can you (or your query planning model ) make from these distributions?

I believe there may be some confusion with how I described uses for the KLL sketches. Let me try to clarify with some more examples.

I used the above `GROUP BY` query above as an example of something that the DB user could write. How they utilize the KLL sketch is entirely up to them. But in implementing support for generating KLL sketches for users in the database, we need to have memory tracking for this scenario, which is why in the implementation of the KLL sketch aggregation function implementation, I had some code similar to the following (referencing my initial response on this PR):

```java
memoryUsage -= sketch.getSerializedSizeBytes();
sketch.update(item);
memoryUsage += sketch.getSerializedSizeBytes();
```

Unfortunately, due to engine design, we can't really special-case function implementations easily for queries with and without these `GROUP BY` statements. If queries don't have `GROUP BY` statements, memory isn't as large of a concern. Most aggregation functions' state rarely exceeds the roughly the 1-megabyte range, so poor memory utilization estimates have a smaller probability of causing issues. However, because of the necessity to implement the memory size tracking, it caused poor performance for queries even without a `GROUP BY` clause. Since the root cause of the poor performance was `KllItemsSketch#getSerializedSizeBytes`, I was motivated to open this PR.

The hypothetical scenario of a large `GROUP BY` query described above is a little different from how we (the DB engine authors) intend to use the KLL sketches. We aren't doing `GROUP BY` queries when we try to generate statistics for a table, so the amount of memory required for the aggregation is far lower and less of a concern, but the query runtime performance is. Essentially generating statistics for a table amounts to running a query like

```sql
SELECT kll_sketch(<column1>), kll_sketch(<column2>), ... FROM <table>
```

The resulting sketches from the query are stored in a location where they can be queried later by the optimizer. Suppose a user comes along and writes the query

```sql
SELECT * from <table> WHERE <column1> <= 5
```

The optimizer will analyze the query and generate a "Filter" operator representing `<column1> <= 5`. The optimizer attempts to estimate the output row count from all nodes in the SQL query plan. For the "Filter" operator, the optimizer will get the sketch stored for `<column1>` and use it to compute the proportion of rows that would be returned by the table scan which satisfy the filter `<= 5`.

To motivate the reason why we want accurate filter estimates, take the following scenario: a user has a very complex query with multiple (>2) `JOIN` statements. You can improve query performance by re-arranging the order which tables are joined together by joining tables with smaller result sets earlier. So, having an accurate estimate of the number of records that satisfy the filter on a table scan can significantly improve query runtime.

leerho:
Thanks Zac, this is super helpful.  I was not aware that you offer sketches for use by your DB-users.  That is wonderful!   Now I realize you have two use cases: one for User Queries (UQ), and one for Query Planning and Optimization (QPO) used by the DB developers.  

Is it fair to assume that the QPO task is always performed prior to any UQ? In other words, you have the opportunity to learn a lot about the user's tables prior to the user using them.

If this is true, then from the QPO pass, you could have in advance of any UQ the **_distribution of element sizes_** of any column the user might select.  (I would hope that this kind of table metadata could be constructed as a table is being built, but that is a different discussion.) 

As I have shown you in several graphs, the average size of KLL sketch is quite predictable from _**k**_ the configured sketch size, **_n_** the total number of elements fed to the sketch and **_s_** the size of the elements fed to the sketch.  If _**s**_ is a random variable, then all we need is the CDF.   Most importantly, we can establish upper bounds on both of those values.  The most conservative approach would be to set _**n**_ to be the number of rows of the entire table and _**s**_ to be the largest element in the column of interest.  From that we can predict pretty accurately (and very quickly) what the size of the sketch will be.   Or, if we discover from experience, that this size is too conservative, from the floats sketch that computes the CDF of size, we could choose the 90th percentile size instead of the max size, etc. 

In other words, instead of trying to track the size of the sketch before and after each update, which is hugely expensive, can we compute a memory budget for the sketch to grow in and use that for memory planning? 

One more question I wanted to ask you is what Java version are you using?  We are in the planning stages to move to Java 17 and 21, but we will have to draw a hard line in the sand for Java 17.  In other words, at a specific DataSketches Version (perhaps version 7 or 8), Java 17+ will be required.  If you will still need Java versions < 17, you will have to use earlier versions of the library.  How will this impact you?

Cheers,
Lee.

ZacBlanco:
> Is it fair to assume that the QPO task is always performed prior to any UQ? In other words, you have the opportunity to learn a lot about the user's tables prior to the user using them.

It's not necessarily the case that we always have the statistics for QPO before UQ, though we recommend most users spend the resources to do so as it can make the queries execute quicker. It depends on the use case though. I think we would be okay with a heuristic.

I think your suggestion of calculating a bound for the memory would work for us. I think collecting a sketch of data sizes alongside the actual sketch could also be a feasible solution that would give us the memory bounds even if the sketch size is unknown and would be less overhead than the current approach. I appreciated you sharing the graphs of the sketch size for a given N. However, I couldn't identify a given formula in the KLL paper[^1] which corresponds to the same types of curves you shared. Could you help point me in the right direction so I could incorporate them?

Given that I can compute a rough bound on the size of the sketch, I think that parts of this PR would not be necessary. However, I do believe that the addition to the `ArrayOfItemsSerde` interface of `isFixedWidth` would still be beneficial for `KllItemsSketch` to improve the performance of `getSerializedSizeBytes()`. I can modify the PR to include only that part if you agree with that.

> One more question I wanted to ask you is what Java version are you using? We are in the planning stages to move to Java 17 and 21, but we will have to draw a hard line in the sand for Java 17. In other words, at a specific DataSketches Version (perhaps version 7 or 8), Java 17+ will be required. If you will still need Java versions < 17, you will have to use earlier versions of the library. How will this impact you?

Currently our build actually relies on Java 8. We are moving to Java 11 soon. Personally, I would like our project to move to more modern versions, but I don't have enough influence to make that happen. It's also complex decision for us to make as many companies use [our system](https://github.com/prestodb/presto). If you move Java versions upwards we would either rely on patch builds to a version before 17, or would just wait until we move to a later Java version to incorporate new versions of the datasketches library.

[^1]: https://arxiv.org/pdf/1603.05346v2

leerho:
The part of the ds-java library that is Java version sensitive is the ds-memory component, which we use for off-heap management.  And currently ds-memory is supported for Java 8 and 11.  With Java 17, there is no longer a need for our Memory component as Java 17 incorporates the new Panama FFM capability as a preview (or incubation).  The FFM will be far more capable and faster, since it leverages native code and built right into the language.  

With Java 17 for the ds-java library we are also going to break up our library into Java Modules (JPMS) as well as move the build tools to Gradle. This means you will only need to download the specific sketches you need which will be in separate jars.

The Gradle build shouldn't impact you unless you wish to build from source.  Also the JPMS shouldn't bother you since you can place all these modular jars on the classpath (instead of the module path) and everything should just work  😄.

WRT the PR.  I am still reviewing your code and I agree that the "isFixedWidth" or isItemFixedSize" (might be a little clearer) is a good contribution.  But you might have a few more nuggets in there.

WRT the graph.  You won't find any formula for producing that graph in the paper.  But it can be inferred from how the sketch grows, which is discussed in the paper.  Building the graph is rather involved and uses code deep in the bowels of the sketch.  I have written a utility that produces it for me, but it is pretty much for my own research.  In your case, I don't think you want the graph, per se, I think what you want is the ability to compute the expected size based on K, n, and item size.  That could be an improvement in the current getMaxSerializedSizeBytes(...) and make it more general.

leerho:
The one concept that I want you to take away from our discussion about KLL sketch size is this:  When _**n**_ gets large enough, the internal sketch structure reaches a stable form. This is where you see the curve slope dramatically flatten. From that point on, as _**n**_  continues to grow the sketch size will increase by adding only 8 items to its retained items for every factor of 2 growth in _**n**_!   For example, looking at the last graph, the stable form is reached at about 100K 64B items.  So if n grows from 100k to 200k items the sketch size increases by about 8 items.  -- Or, growing from 1T items to 2T items the sketch increases in size by about 8 items!  Thus, planning for a memory space for the sketch to grow in should be straight-forward.

ZacBlanco:
When you say "direct mode", I'm interpreting that as you referring to `KllDoublesSketch` rather than `KllItemsSketch<Double>`. If I am correct with that assumption, then yes I think we will be using them. After benchmarking, I'm convinced the performance benefit is going to be worth the additional implementation complexity on our side, so I would like to use them. Hence why I opened #556. That `long` version of the sketch should be the only other implementation that we will need as most numeric data that benefits from the sketches in the initial implementation is either `double` or `long`.

leerho:
Through our discussion, Zac has decided he is willing to try another approach.  I will close this PR now, but it can always be reopened.

PR #555, opened 2024-04-29
Fix pom xml header

leerho:
According to Maven documentation [https://maven.apache.org/pom.html#The_Basics](https://maven.apache.org/pom.html#The_Basics), only the xsd URI can be https. the other 3 URIs in the header have to be http.

Only XML validators will pick this up. So I don't think this warrants a bug fix.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,9 +18,9 @@ specific language governing permissions and limitations
 under the License.
 -->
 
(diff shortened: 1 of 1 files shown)


leerho:
Sorry, forgot to push!

jmalkin (approved):

PR #556, opened 2024-05-01
Add native longs KLL sketch

ZacBlanco:
Our use case would benefit from an implementation for this native type. I considered a commit to refactor some common logic, but found little benefit, so opted to essentially `sed s/float/long/` the `KllFloatsSketch` implementation along with all of the associated classes and tests.


leerho:
I haven't had a chance to review this carefully, but I will.  In principle I am in favor of this.

jmalkin:
It's not an elegant solution, but effectively doing `s/float/long/g` is, sadly, a fairly reasonable approach in Java.

leerho:
Merge into branch "kll_longs_sketch"

PR #561, opened 2024-05-23
Kll longs sketch

leerho:
In the main code, nearly all the corrections are in the Javadocs.

However, in the test code, there were many places where he was comparing a long vs a float ('cause he copied the code from the float classes).

Also, he didn't check the test classes with printing enabled. There were
numerous places where the printf format expected an "%f" that should have been a "%d".

Nonetheless, Zac did an amazing job of creating this KllLongsSketch, considering the complexity of the code!

I also did a coverage tests and made sure that the coverage of the "longs" classes were very close to the coverage of the comparable "doubles" and "floats" classes -- and they were!  This means that he was able to faithfully translate all the relevant tests from float to longs and I did not find any gaps in testing.

### DIFF
--- a/src/main/java/org/apache/datasketches/kll/KllDirectLongsSketch.java
+++ b/src/main/java/org/apache/datasketches/kll/KllDirectLongsSketch.java
@@ -19,17 +19,36 @@
 
 package org.apache.datasketches.kll;
 
+import static org.apache.datasketches.common.ByteArrayUtil.copyBytes;
+import static org.apache.datasketches.kll.KllPreambleUtil.DATA_START_ADR;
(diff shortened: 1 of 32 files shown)


leerho:
I admit a few of these files are KLL Doubles Sketches where my search string picked up the same errors in those files, so I just corrected them and included them here.  Going forward I will create a separate PR for these other extraneous errors.

leerho:
This is difficult to review because this view does not show the context.
- To see the changes I made to Zac's code you will need to compare this branch at commit d41aafc against this branch at commit 201869e.
- Another way to  review this is to compare this code compared to sister classes like the relevant classes from the KllDoublesSketch. They should line up exactly except for the difference in the primitive, "double" verses "long".  I use a separate application for this.

AlexanderSaydakov (approved):

PR #566, opened 2024-05-31
fixed find_run_start() and removed adjustment in insert_new_run()

AlexanderSaydakov:


c-dickens (approved):
Looks good.  I understand the off-by-one problem that you mentioned now.

PR #572, opened 2024-06-10
Optimise HLL union for mergeHLLtoHLL

freakyzoidberg:
This PR introduce JMH benchmark - happy to remove it and leave that to characterisation if thats an issue.

This also only shows in a particular place that some optimisation can be gained by avoiding a call to Math.max + array mutation with a simple comparaison.

About 8% gain for HLL heap union

Before
```
Benchmark                                       Mode  Cnt    Score    Error  Units
BenchmarkUnion.benchHllUnionHLLToHLLMerge  avgt    5  511.411 ± 12.468  ms/op
```

After
```
Benchmark                                       Mode  Cnt    Score   Error  Units
BenchmarkUnion.benchHllUnionHLLToHLLMerge  avgt    5  471.329 ± 9.712  ms/op
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,6 +89,7 @@ under the License.
 
     <!-- Test -->
     <testng.version>7.5.1</testng.version>
+    <jmh.version>1.37</jmh.version>
     <!-- these are TestNG groups used for excluding / including groups of tests. See profiles section. -->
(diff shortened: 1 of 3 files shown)


AlexanderSaydakov:
I vaguely remember when I worked on similar improvements I found that max is faster than comparison. My understanding is that branch prediction misses are expensive, and max can be implemented with no branches.

PR #573, opened 2024-06-13
style changes

AlexanderSaydakov:
no changes in behavior, style only (indentation, naming)


leerho (approved):
Modern Java styling no longer recommends using a trailing (or leading) "_" character for class member variables.  It used to be fairly common years ago but you won't see it in more recent JVM java classes.  Most IDEs will emphasize class members with bold or colored text.  It is not wrong, but code using this convention tends to look "dated" :)

PR #575, opened 2024-06-18
expansion and some more cleanup

AlexanderSaydakov:


src/main/java/org/apache/datasketches/filters/quotientfilter/QuotientFilter.java
```diff
@@ -21,46 +21,34 @@
...
 
+import org.apache.datasketches.common.SketchesException;
 import org.apache.datasketches.filters.common.BitArray;
 import org.apache.datasketches.filters.common.HeapBitArray;
 
 public class QuotientFilter extends Filter {
 
-  int bitPerEntry_;
-  int fingerprintLength_;
+  public static final double LOAD_FACTOR = 0.9;
```
leerho, line 34: Based on our discussion this morning, I think having a DEFAULT_LOAD_FACTOR and then a way for the user to set a different load factor.  Based on characterization studies, the default should probably be a little lower.

leerho (approved):

PR #588, opened 2024-08-17
Fix a memory leak in test.

leerho:
This was a memory leak in the tests.  It only shows up when the tests run long enough and generate enough garbage for the garbage collector to activate and call finalize.  It only showed up when running all the tests for the java library.  Running individual tests it will not show up. I would see the warning after all the tests have run, but, of course, the warning message doesn't provide any clues about where the culprit  is, it can't!  It was kind of a cat-and-mouse game to find it, but I finally figured out a way to track it down.  

This leak has been there a while, but it was only in the test code. So no real harm done.  Release 6.1.0 is coming soon, I think it is safe to release it then.

### DIFF
--- a/src/test/java/org/apache/datasketches/theta/UnionImplTest.java
+++ b/src/test/java/org/apache/datasketches/theta/UnionImplTest.java
@@ -191,8 +191,8 @@ public class UnionImplTest {
     final int k = 1 << 12;
     final int u = 2 * k;
     final int bytes = Sketches.getMaxUpdateSketchBytes(k);
-    WritableMemory wmem = WritableMemory.allocateDirect(bytes / 2); //too small, forces new allocation on heap
-    WritableMemory wmem2 = WritableMemory.allocateDirect(bytes / 2);
+    WritableMemory wmem = WritableMemory.allocateDirect(bytes / 2); //not really used, except as a reference.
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/datasketches/theta/UnionImplTest.java
```diff
@@ -191,21 +191,21 @@ public void checkMoveAndResizeOffHeap() {
...
     for (int i = 0; i < u; i++) { union.update(i); }
-    assertFalse(union.isSameResource(wmem));
+    assertFalse(union.isSameResource(wmem)); //different Memories altogether
 
     final Union union2 = SetOperation.builder().buildUnion(); //on-heap union
     assertFalse(union2.isSameResource(wmem2));  //obviously not
-    wmem.close();
-    //note wmem2 has already been closed by the DefaultMemoryRequestServer
+    wmem.close(); //empty, but we must close it anyway.
+    //note wmem2 has already been closed by the DefaultMemoryRequestServer.
```
jmalkin, line 208: could do `assertFalse(wmem2.isAlive())` here rather than just a comment?

leerho: Fixed.

jmalkin (approved):

PR #617, opened 2024-12-17
Pull master into 6.1.X

leerho:
Once I create a 'main' branch it will initially be based on J-17. After that point, any further fixes to the Java 8-11 version of ds-java in the 6.1.X branch will have to be developed in 6.1.X branch, thus making the 6.1.X effectively the development branch for Java 8-11.  

I am having difficulty getting different opinions from git and Eclipse as to whether these to branches are identical or not.  Eclipse tells me that they are now identical - which they should be.  But 'git difftool ...' says there are differences!
I'm not sure what to believe.  

So I want to merge them to make sure they are the same.  These tools are impacting my sanity!


leerho:
This is a defective PR and shows false differences between files that are identical.

PR #621, opened 2024-12-19
fix indentations in all workflows and try to fix CodeQL

leerho:


leerho:
The CodeQL workflow is not able to automatically build and so I am going to have to substitute a custom build into the workflow and see if that fixes the problem.

AlexanderSaydakov (approved):

PR #633, opened 2025-01-14
remove pattern that seems to be problematic

AlexanderSaydakov:


leerho:
That is what I guessed.  But we have to remove this offending pattern from the other workflows as well.

leerho (approved):

PR #639, opened 2025-01-21
Changed java '1.8' to Java '8'

leerho:
I didn't realize that the Java_setup workflows do not accept the specification 'java 1.8'. They only accept 'java 8' !!
This 2 character change is required in all 5 workflows.


leerho:
I double checked these workflows on my personal repo and they all work.

jmalkin (approved):

PR #640, opened 2025-01-21
Fix target directory in workflow back to /site/

leerho:
I thought this was working, but now it wants to deploy from /target/site/apidocs, when before it was /target/reports/apidocs!


jmalkin (approved):
I feel like we already did this change. I wonder if it got reverted accidentally. But I'm also unlikely to spend time investigating that.

PR #644, opened 2025-01-26
use iteration in set ops, wrap compressed sketch and unpack in iterator

AlexanderSaydakov:


AlexanderSaydakov:
This should speed up union of wrapped compressed Theta sketches

src/main/java/org/apache/datasketches/theta/IntersectionImpl.java
```diff
@@ -448,27 +446,16 @@ private void performIntersect(final Sketch sketchIn) {
...
-        final long hashIn = cacheIn[i];
-        if (hashIn <= 0L || hashIn >= thetaLong_) { continue; }
-        final int foundIdx = hashSearch(hashTable, lgArrLongs_, hashIn);
-        if (foundIdx == -1) { continue; }
-        matchSet[matchSetCount++] = hashIn;
+        if (foundIdx != -1) {
+          matchSet[matchSetCount++] = hashIn;
+        }
+      } else {
+        if (sketchIn.isOrdered()) { break; } // early stop
```
jmalkin, line 458: This won't change over the course of the intersection. Should it be a flag pre-computed in advance to avoid a function call every time? Not sure if the JVM can easily determine there's no chance of a side-effect from other code.

src/main/java/org/apache/datasketches/theta/DirectCompactCompressedSketch.java
```diff
@@ -0,0 +1,134 @@
...
+ * <p>This sketch can only be associated with a Serialization Version 4 format binary image.</p>
+ *
+ * <p>This implementation uses data in a given Memory that is owned and managed by the caller.
+ * This Memory can be off-heap, which if managed properly will greatly reduce the need for
+ * the JVM to perform garbage collection.</p>
+ */
+class DirectCompactCompressedSketch extends DirectCompactSketch {
+  /**
+   * Construct this sketch with the given memory.
+   * @param mem Read-only Memory object with the order bit properly set.
```
jmalkin, line 45: Rather than "properly set" maybe just say it must be ordered and with the bit set to true?

src/main/java/org/apache/datasketches/theta/DirectCompactCompressedSketch.java
```diff
@@ -0,0 +1,134 @@
...
+   */
+  DirectCompactCompressedSketch(final Memory mem) {
+    super(mem);
+  }
+
+  /**
+   * Wraps the given Memory, which must be a SerVer 4 compressed CompactSketch image.
+   * Must check the validity of the Memory before calling. The order bit must be set properly.
```
jmalkin, line 53: again

jmalkin (commented):
I did not quite get the logic in DirectCompactCompressedSketch.getRetainedEntries but there were a few other things that seemed worth noting.

jmalkin:
Ok, for some reason I was thinking it was going over the entries themselves, not the value representing the number of entries. Got it now and it's fine.

src/main/java/org/apache/datasketches/theta/UnionImpl.java
```diff
@@ -328,38 +327,13 @@ public void union(final Sketch sketchIn) {
...
-          gadget_.hashUpdate(hashIn); //backdoor update, hash function is bypassed
-          c++; //ensures against invalid state inside the incoming sketch
-        }
+    HashIterator it = sketchIn.iterator();
+    while (it.next()) {
+      final long hash = it.get();
+      if (hash < unionThetaLong_ && hash < gadget_.getThetaLong()) {
+        gadget_.hashUpdate(hash); // backdoor update, hash function is bypassed
+      } else {
+        if (sketchIn.isOrdered()) { break; }
```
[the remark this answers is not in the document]
jmalkin: The comment on intersection appeared higher in my review. So the longer version of the relevant comment is that one.

src/main/java/org/apache/datasketches/theta/IntersectionImpl.java
AlexanderSaydakov: Generally I prefer not to introduce aliases, but if you think this can help, I can do that. Perhaps we could try measuring the effect one day.

src/main/java/org/apache/datasketches/theta/DirectCompactCompressedSketch.java
AlexanderSaydakov: I think this refers to memory byte order. Perhaps can be rephrased for clarity.

AlexanderSaydakov:
If it was a bit difficult and confusing, perhaps some comments are needed. Let me see what I can do.

jmalkin (approved):
Looks good

PR #657, opened 2025-03-29
Main21-2 to main

leerho:
All but one of these changes are Javadoc changes to satisfy new Javadoc checks introduced with Java 21. 
The new requirements are that all public methods and classes must have a Javadoc description.  A new twist is that public static classes that don't really need a constructor, must have a constructor with a description!

### DIFF
--- a/src/main/java/org/apache/datasketches/cpc/MergingValidation.java
+++ b/src/main/java/org/apache/datasketches/cpc/MergingValidation.java
@@ -56,7 +56,7 @@ public class MergingValidation {
   private PrintWriter printWriter;
 
   /**
-   *
+   * Constructor
    * @param lgMinK lgMinK
    * @param lgMaxK lgMaxK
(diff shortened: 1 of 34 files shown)


src/main/java/org/apache/datasketches/cpc/MergingValidation.java
```diff
@@ -79,7 +79,7 @@ public MergingValidation(final int lgMinK, final int lgMaxK, final int lgMulK, f
   }
 
   /**
-   *
+   * Start the merging validation process
    */
   public void start() {
```
jmalkin, line 84: That this class exists in the main code at all is kind of ugly, imho. Sigh.

jmalkin (approved):
I do wish we'd find a better solution for that CPC test code than to stick it in main. But not gonna condition approving this on solving that.

PR #658, opened 2025-03-31
Update 8.0.X POM to 8.0.0 (remove -SNAPSHOT).

leerho:


jmalkin:
This would merge the change into main, which I don't think is what we want?

PR #666, opened 2025-05-12
Add Go cross-serde tests for CPC sketch 

QasimKhan5d:
Added cross language deserialization tests for serialized sketches produced by datasketches-go

PR #676, opened 2025-07-24
CMS and CPCxLangTest move to FFM, Build fix 

freakyzoidberg:
Fix build following conflict introduced by concurrent change in https://github.com/apache/datasketches-java/pull/667 and https://github.com/apache/datasketches-java/pull/666

This pull request introduces uniformisation with the rest of the library to the `CountMinSketch` implementation, including improved serialization, memory efficiency, and validation checks. It also updates test cases to reflect these changes. The most important changes are grouped below by theme:

### Core Improvements to `CountMinSketch`:
* Added a thread-local `MemorySegment` for efficient long-to-byte conversions, reducing allocations in hot paths (`src/main/java/org/apache/datasketches/count/CountMinSketch.java`, [[1]](diffhunk://#diff-a27b2c7ae95edb924c40143bf39f23acc3cab7f6a03b674262849f067b6ac2d6R57-R59) [[2]](diffhunk://#diff-a27b2c7ae95edb924c40143bf39f23acc3cab7f6a03b674262849f067b6ac2d6L60-R135).
* Introduced robust validation for `numHashes` and `numBuckets` parameters, including checks for overflow and better error messages (`src/main/java/org/apache/datasketches/count/CountMinSketch.java`, [src/main/java/org/apache/datasketches/count/CountMinSketch.javaL60-R135](diffhunk://#diff-a27b2c7ae95edb924c40143bf39f23acc3cab7f6a03b674262849f067b6ac2d6L60-R135)).
* Replaced `ByteBuffer` usage with `MemorySegment` for serialization/deserialization, improving performance and simplifying code (`src/main/java/org/apache/datasketches/count/CountMinSketch.java`, [src/main/java/org/apache/datasketches/count/CountMinSketch.javaL387-R492](diffhunk://#diff-a27b2c7ae95edb924c40143bf39f23acc3cab7f6a03b674262849f067b6ac2d6L387-R492)).
* Added a new `toByteArray` method for serialization and refactored the deserialization logic for better maintainability (`src/main/java/org/apache/datasketches/count/CountMinSketch.java`, [[1]](diffhunk://#diff-a27b2c7ae95edb924c40143bf39f23acc3cab7f6a03b674262849f067b6ac2d6L345-R438) [[2]](diffhunk://#diff-a27b2c7ae95edb924c40143bf39f23acc3cab7f6a03b674262849f067b6ac2d6L387-R492).

### Code Simplifications and Optimizations:
* Refactored methods like `getHashes`, `update`, and `getEstimate` to use `MemorySegment` for byte array conversions and improved readability with consistent use of `final` variables (`src/main/java/org/apache/datasketches/count/CountMinSketch.java`, [[1]](diffhunk://#diff-a27b2c7ae95edb924c40143bf39f23acc3cab7f6a03b674262849f067b6ac2d6L174-R221) [[2]](diffhunk://#diff-a27b2c7ae95edb924c40143bf39f23acc3cab7f6a03b674262849f067b6ac2d6L202-R249) [[3]](diffhunk://#diff-a27b2c7ae95edb924c40143bf39f23acc3cab7f6a03b674262849f067b6ac2d6L214-R260) [[4]](diffhunk://#diff-a27b2c7ae95edb924c40143bf39f23acc3cab7f6a03b674262849f067b6ac2d6L242-R291).
* Simplified loop logic in methods like `getEstimate` to avoid redundant processing (`src/main/java/org/apache/datasketches/count/CountMinSketch.java`, [src/main/java/org/apache/datasketches/count/CountMinSketch.javaL242-R291](diffhunk://#diff-a27b2c7ae95edb924c40143bf39f23acc3cab7f6a03b674262849f067b6ac2d6L242-R291)).

### Test Case Updates:
* Updated test cases to use the new `toByteArray` method for serialization and adjusted assertions for compatibility with the refactored deserialization logic (`src/test/java/org/apache/datasketches/count/CountMinSketchTest.java`, [[1]](diffhunk://#diff-c921c46151597199d049e4cd7ad310ae129befa8b0f9eadddd469c8708c965c5L206-R206) [[2]](diffhunk://#diff-c921c46151597199d049e4cd7ad310ae129befa8b0f9eadddd469c8708c965c5L231-R231).
* Modified cross-language test in `CpcSketchCrossLanguageTest` to use `MemorySegment` for consistency with the updated serialization approach (`src/test/java/org/apache/datasketches/cpc/CpcSketchCrossLanguageTest.java`, [src/test/java/org/apache/datasketches/cpc/CpcSketchCrossLanguageTest.javaL92-R92](diffhunk://#diff-74cb379136cdd3e590dd47ceed651a002e8cc2d1a581adb6260f4dc065140b75L92-R92)).

### DIFF
--- a/src/main/java/org/apache/datasketches/count/CountMinSketch.java
+++ b/src/main/java/org/apache/datasketches/count/CountMinSketch.java
@@ -31,9 +31,9 @@ import java.nio.charset.StandardCharsets;
 import static java.lang.foreign.ValueLayout.JAVA_BYTE;
 import java.util.Random;
 
-import static org.apache.datasketches.common.SpecialValueLayouts.JAVA_INT_UNALIGNED_BIG_ENDIAN;
-import static org.apache.datasketches.common.SpecialValueLayouts.JAVA_LONG_UNALIGNED_BIG_ENDIAN;
-import static org.apache.datasketches.common.SpecialValueLayouts.JAVA_SHORT_UNALIGNED_BIG_ENDIAN;
+import static java.lang.foreign.ValueLayout.JAVA_INT_UNALIGNED;
+import static java.lang.foreign.ValueLayout.JAVA_LONG_UNALIGNED;
+import static java.lang.foreign.ValueLayout.JAVA_SHORT_UNALIGNED;
 
 
 public class CountMinSketch {
@@ -113,7 +113,7 @@ public class CountMinSketch {
    */
   private static byte[] longToBytes(final long value) {
     final MemorySegment segment = LONG_SEGMENT.get();
-    segment.set(JAVA_LONG_UNALIGNED_BIG_ENDIAN, 0, value);
+    segment.set(JAVA_LONG_UNALIGNED, 0, value);
     return segment.toArray(JAVA_BYTE);
   }
 
@@ -393,9 +393,9 @@ public class CountMinSketch {
   public byte[] toByteArray() {
     final int serializedSizeBytes = getSerializedSizeBytes();
     final MemorySegment wseg = MemorySegment.ofArray(new byte[serializedSizeBytes]);
-    
+
     long offset = 0;
-    
+
     // Long 0
     final int preambleLongs = Family.COUNTMIN.getMinPreLongs();
     wseg.set(JAVA_BYTE, offset++, (byte) preambleLongs);
@@ -406,31 +406,31 @@ public class CountMinSketch {
     final int flagsByte = isEmpty() ? Flag.IS_EMPTY.mask() : 0;
     wseg.set(JAVA_BYTE, offset++, (byte) flagsByte);
     final int NULL_32 = 0;
-    wseg.set(JAVA_INT_UNALIGNED_BIG_ENDIAN, offset, NULL_32);
+    wseg.set(JAVA_INT_UNALIGNED, offset, NULL_32);
     offset += 4;
 
     // Long 1
-    wseg.set(JAVA_INT_UNALIGNED_BIG_ENDIAN, offset, numBuckets_);
+    wseg.set(JAVA_INT_UNALIGNED, offset, numBuckets_);
     offset += 4;
     wseg.set(JAVA_BYTE, offset++, numHashes_);
     short hashSeed = Util.computeSeedHash(seed_);
-    wseg.set(JAVA_SHORT_UNALIGNED_BIG_ENDIAN, offset, hashSeed);
+    wseg.set(JAVA_SHORT_UNALIGNED, offset, hashSeed);
     offset += 2;
     final byte NULL_8 = 0;
     wseg.set(JAVA_BYTE, offset++, NULL_8);
-    
+
     if (isEmpty()) {
       return wseg.toArray(JAVA_BYTE);
     }
 
-    wseg.set(JAVA_LONG_UNALIGNED_BIG_ENDIAN, offset, totalWeight_);
+    wseg.set(JAVA_LONG_UNALIGNED, offset, totalWeight_);
     offset += 8;
 
     for (long w: sketchArray_) {
-      wseg.set(JAVA_LONG_UNALIGNED_BIG_ENDIAN, offset, w);
+      wseg.set(JAVA_LONG_UNALIGNED, offset, w);
       offset += 8;
     }
-    
+
     return wseg.toArray(JAVA_BYTE);
   }
 
@@ -448,13 +448,13 @@ public class CountMinSketch {
     final byte serialVersion = buf.get(JAVA_BYTE, offset++);
     final byte familyId = buf.get(JAVA_BYTE, offset++);
     final byte flagsByte = buf.get(JAVA_BYTE, offset++);
-    final int NULL_32 = buf.get(JAVA_INT_UNALIGNED_BIG_ENDIAN, offset);
+    final int NULL_32 = buf.get(JAVA_INT_UNALIGNED, offset);
     offset += 4;
 
-    final int numBuckets = buf.get(JAVA_INT_UNALIGNED_BIG_ENDIAN, offset);
+    final int numBuckets = buf.get(JAVA_INT_UNALIGNED, offset);
     offset += 4;
     final byte numHashes = buf.get(JAVA_BYTE, offset++);
-    final short seedHash = buf.get(JAVA_SHORT_UNALIGNED_BIG_ENDIAN, offset);
+    final short seedHash = buf.get(JAVA_SHORT_UNALIGNED, offset);
     offset += 2;
     final byte NULL_8 = buf.get(JAVA_BYTE, offset++);
 
@@ -468,12 +468,12 @@ public class CountMinSketch {
     if (empty) {
       return cms;
     }
-    long w = buf.get(JAVA_LONG_UNALIGNED_BIG_ENDIAN, offset);
+    long w = buf.get(JAVA_LONG_UNALIGNED, offset);
     offset += 8;
     cms.totalWeight_ = w;
 
     for (int i = 0; i < cms.sketchArray_.length; i++) {
-      cms.sketchArray_[i] = buf.get(JAVA_LONG_UNALIGNED_BIG_ENDIAN, offset);
+      cms.sketchArray_[i] = buf.get(JAVA_LONG_UNALIGNED, offset);
       offset += 8;
     }
 
--- a/src/main/java/org/apache/datasketches/count/CountMinSketch.java
+++ b/src/main/java/org/apache/datasketches/count/CountMinSketch.java
@@ -36,6 +36,16 @@ import static java.lang.foreign.ValueLayout.JAVA_LONG_UNALIGNED;
 import static java.lang.foreign.ValueLayout.JAVA_SHORT_UNALIGNED;
 
 
+/**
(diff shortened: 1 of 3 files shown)


jmalkin:
Not thrilled with the use of bytebuffer and its endian quirks. I'd look for a way to do it with memory/FFM

src/main/java/org/apache/datasketches/count/CountMinSketch.java
```diff
@@ -22,8 +22,8 @@
 import org.apache.datasketches.common.Family;
 import org.apache.datasketches.common.SketchesArgumentException;
 import org.apache.datasketches.common.SketchesException;
+import org.apache.datasketches.common.Util;
 import org.apache.datasketches.hash.MurmurHash3;
-import org.apache.datasketches.tuple.Util;
 
 import java.io.ByteArrayOutputStream;
 import java.nio.ByteBuffer;
```
leerho: I do not recommend the use of ByteBuffer or any of its subclasses.  Since the main branch is targeted for Java 25, I recommend switching to use FFM MemorySegments. Here are my reasons:
- ByteBuffer off-heap allocations cannot be interactively (dynamically) closed, they are managed by the GC.  Thus out-of-memory scenarios can happen in large systems with millions of sketches because the GC can't keep up with new off-heap allocations.
- The default endianness of BB is Big-Endian!  Even if you intentionally set it to Little-Endian, it will silently switch back to BE after duplicate(), clear(), or slice(), operations.  To make matters worse, this silent behavior of BB is not documented in the official Javadocs!  
- Managing endianness with BB requires a great deal of care and it is very easy to forget to constantly switch the endianness back to LE.  Thus, I consider its use here in this class as dangerous since there is no explicit settings of endianness, (whether to LE or BE!).  
- Endianness is explicitly set with FFM (and with the older datasketches.memory classes) with no silent switching of endianness, thus much safer.
- Our library is completely LE with one exception in t-Digest for binary compatibility with externally produced t-Digest sketches.  Since the introduction of CountMin is new for the Java library it should use LE ordering, since, I assume, there is no history of CountMin sketches being stored.
- Our Theta, Tuple and CPC sketches must be LE due to their compression algorithms, which rely on LE byte ordering.

src/main/java/org/apache/datasketches/count/CountMinSketch.java
```diff
@@ -22,8 +22,8 @@
 import org.apache.datasketches.common.Family;
 import org.apache.datasketches.common.SketchesArgumentException;
 import org.apache.datasketches.common.SketchesException;
+import org.apache.datasketches.common.Util;
 import org.apache.datasketches.hash.MurmurHash3;
-import org.apache.datasketches.tuple.Util;
```
leerho, line 26: This reference to /tuple.Util is only used for the computeSeedHash() in lines 364, 402 and 404.  Because this method and the companion method checkSeedHashes() are used in other sketches and not just the Tuple Sketch, these two functions have been moved to /datasketches.common.Util. in the current main. All you need to do is change this import to .../datasketches.common.Util.

src/test/java/org/apache/datasketches/cpc/CpcSketchCrossLanguageTest.java
```diff
@@ -89,7 +89,7 @@ public void checkAllFlavorsGo() throws IOException {
     int flavorIdx = 0;
     for (int n: nArr) {
       final byte[] bytes = Files.readAllBytes(goPath.resolve("cpc_n" + n + "_go.sk"));
-      final CpcSketch sketch = CpcSketch.heapify(Memory.wrap(bytes));
+      final CpcSketch sketch = CpcSketch.heapify(MemorySegment.ofArray(bytes));
       assertEquals(sketch.getFlavor(), flavorArr[flavorIdx++]);
```
leerho, line 93: Thanks for catching this!

freakyzoidberg:
Very fair - Updating CMS to remove ByteBuffer family altogether follwing the same pattern as the FFM current work effort - 

Thanks - sorry should have caught that earlier..

freakyzoidberg:
There it should be more aligned - now I dont think those are forcing the endian to LE - rather being nativeOrder/platform dependant.

If this is really an issue - that should be addresse widely through all the FFM assumed to be LE

That PR actually fix the build of the main branch 

cc @leerho

src/main/java/org/apache/datasketches/count/CountMinSketch.java
```diff
@@ -406,31 +406,31 @@ public byte[] toByteArray() {
     final int flagsByte = isEmpty() ? Flag.IS_EMPTY.mask() : 0;
     wseg.set(JAVA_BYTE, offset++, (byte) flagsByte);
     final int NULL_32 = 0;
-    wseg.set(JAVA_INT_UNALIGNED_BIG_ENDIAN, offset, NULL_32);
+    wseg.set(JAVA_INT_UNALIGNED, offset, NULL_32);
     offset += 4;
```
leerho: This (and following) is a use-case where the new datasketches.common.positional.PositionalSegment could be used.  You might want to look at it. I used the term "positional" instead of "buffer", because "buffer" is a widely used generic term and doesn't convey what is actually going on.  The PS would eliminate all of the offset tracking that you had to do by hand (with magic numbers).  

I had already migrated theta and tuple, but when I got to bloomfilter, I decided there was a real need for it -- so I paused and created it. It is already used in bloomfilter, tdigest, kll, and req.  I intend to go back and use it in the rest of the library where it makes sense.  And here it makes perfect sense, but this suggestion is optional.

Nonetheless, if you decide not to use PS, I would recommend not using magic numbers for the offset adjustments and use the build-in static constants such as Integer.BYTES, Long.BYTES, etc. The code will become more obvious as to what you are doing and more robust. :-)

Again, thank you for your work on this!

src/main/java/org/apache/datasketches/count/CountMinSketch.java
```diff
@@ -384,20 +441,22 @@ public void serialize(ByteArrayOutputStream buf) {
...
+    final byte flagsByte = buf.get(JAVA_BYTE, offset++);
+    final int NULL_32 = buf.get(JAVA_INT_UNALIGNED, offset);
+    offset += 4;
+
+    final int numBuckets = buf.get(JAVA_INT_UNALIGNED, offset);
+    offset += 4;
+    final byte numHashes = buf.get(JAVA_BYTE, offset++);
+    final short seedHash = buf.get(JAVA_SHORT_UNALIGNED, offset);
+    offset += 2;
+    final byte NULL_8 = buf.get(JAVA_BYTE, offset++);
```
leerho, line 477: The variables preambleLongs, serialVersion, familyId, NULL_32 and NULL_8 are not used.

leerho (changes requested):
A few comments,
1.  Potential use of PositionalSegment
2. Replace magic numbers with built-in constants
3. Why are a bunch of variables created at the end of deserialization if they are not used? There is an opportunity to use these variables to check that they are the correct values -- thus providing confidence that the MemorySegment or byte array image is correct and not garbage.

src/main/java/org/apache/datasketches/count/CountMinSketch.java
```diff
@@ -22,14 +22,19 @@
...
+import static java.lang.foreign.ValueLayout.JAVA_BYTE;
 import java.util.Random;
 
+import static java.lang.foreign.ValueLayout.JAVA_INT_UNALIGNED;
+import static java.lang.foreign.ValueLayout.JAVA_LONG_UNALIGNED;
+import static java.lang.foreign.ValueLayout.JAVA_SHORT_UNALIGNED;
+
 
 public class CountMinSketch {
   private final byte numHashes_;
```
leerho, line 50: Missing Javadoc for public class

src/main/java/org/apache/datasketches/count/CountMinSketch.java
```diff
@@ -57,30 +65,59 @@ int mask() {
...
+    }
 
+    // Validate numBuckets with clear mathematical justification
+    if (numBuckets <= 0) {
+      throw new SketchesArgumentException("Number of buckets must be positive, got: " + numBuckets);
+    }
     if (numBuckets < 3) {
-      throw new SketchesArgumentException("Using fewer than 3 buckets incurs relative error greater than 1.");
+      throw new SketchesArgumentException("Number of buckets must be at least 3 to ensure relative error ≤ 1.0. " +
+          "With " + numBuckets + " buckets, relative error would be " + String.format("%.3f", Math.exp(1.0) / numBuckets));
```
leerho, line 89: The "+" should be at the beginning of the next line.

src/main/java/org/apache/datasketches/count/CountMinSketch.java
```diff
@@ -57,30 +65,59 @@ int mask() {
...
 
+    numHashes_ = numHashes;
+    numBuckets_ = numBuckets;
+    seed_ = seed;
+    hashSeeds_ = new long[numHashes];
+    sketchArray_ = new long[(int) totalSize];
+    totalWeight_ = 0;
+
     Random rand = new Random(seed);
     for (int i = 0; i < numHashes; i++) {
```
leerho, line 116: missing "final" for Random rand...
there are a lot more of these...
Also check for unnecessary casts.

leerho (approved):
Excellent work!

PR #679, opened 2025-07-31
Add test showing HLL merge dependence

freakyzoidberg:
This pull request introduces a new test class, `HllSketchMergeOrderTest`, to investigate and demonstrate the order dependency of DataSketch HLL merge operations. The test highlights that merging HLL sketches in different orders can produce varying cardinality estimates, which has significant implications for applications relying on consistent results.

#### New Test for Merge Order Dependency:
* Added `HllSketchMergeOrderTest` class to demonstrate that merging HLL sketches in different orders (e.g., ABC, CBA, BAC) can lead to different cardinality estimates, proving that the operations are not always commutative or associative. This is especially evident with specific data patterns like powers of 2.

#### Supporting Methods:
* Implemented `createPowersOf2Sketch` method to generate HLL sketches with powers-of-2 values, which are prone to triggering order dependency.
* Added `mergeThreeSketches` method to merge three sketches in a specified order and return the cardinality estimate.

#### Documentation and Findings:
* Included detailed comments and documentation within the test class to explain the findings, key implications, and the mathematical expectations violated by the observed behavior.

### DIFF
--- a/src/test/java/org/apache/datasketches/hll/HllSketchMergeOrderTest.java
+++ b/src/test/java/org/apache/datasketches/hll/HllSketchMergeOrderTest.java
@@ -1,0 +1,116 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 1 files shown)


leerho:
There is no need to merge this. I can pick it up directly from the branch hll-merge-order.

I added a second class: **HllSketchMergeOrderBTest**, which demonstrates that if you use the Composite Estimator, the
sensitivity to merge order disappears completely.

PR #680, opened 2025-08-01
Fix check-go-files to actually enable go test (and not cpp ones)

freakyzoidberg:
This pull request updates the `pom.xml` file to adjust the test groups configuration for the Maven Surefire Plugin. 

* Updated the `<groups>` configuration for check-go-files to use `${testng.check-go-files}` instead of `${testng.check-cpp-files}`. (`[pom.xmlL666-R666](diffhunk://#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L666-R666)`)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -663,7 +663,7 @@ under the License.
               <groupId>org.apache.maven.plugins</groupId>
               <artifactId>maven-surefire-plugin</artifactId>
               <configuration>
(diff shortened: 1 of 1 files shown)


leerho:
This fix is already in the FFM_phase7 PR. So this shouldn't be needed.

PR #681, opened 2025-08-03
Made detection of BigEndian consistent throughout.

leerho:
### Major Goal: Make detection of Big Endian consistent throughout
- Created static initializers in **/common/Family** and **/common/Util** that check for _Native Endianness_ and throws an error if _BigEndian_ is detected.  
- All sketches should use **Family** when serializing, and many classes depend on **/common/Util**.
- This library does not support _BigEndian_ machines for several reasons.
    - Cross platform compatibility
    - Compression algorithms are byte order dependent
    - Hashing is byte order dependent.
 - This involved 
     - Removing unnecessary code that detected BigEndian
     - Redefining bit 0 of the Flags field as "Reserved" and not set or read.

### Minor Goal: cleanup of above files that were touched:
- Added finals (automatically) where possible.
- Added parentheses (automatic) to reduce ambiguity
- Fixed some missing license headers.
- Organized inputs
### Misc
- Added a HllSketch Merge Order Test (based on discussion with @freakyzoidberg )

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -3,7 +3,7 @@ Copyright 2025 The Apache Software Foundation
 
 Copyright 2015-2018 Yahoo Inc.
 Copyright 2019-2020 Verizon Media
-Copyright 2021-     Yahoo Inc.
+Copyright 2021-2025 Yahoo Inc.
 
 This product includes software developed at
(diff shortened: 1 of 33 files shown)


leerho:
I reviewed the Copilot comments.  I do not recommend any changes.

PR #682, opened 2025-08-04
Fix get result and reset

leerho:
This fixes the "getResultAndReset" issue flagged by Copilot, but does it the right way :)

### DIFF
--- a/src/main/java/org/apache/datasketches/quantiles/DirectUpdateDoublesSketch.java
+++ b/src/main/java/org/apache/datasketches/quantiles/DirectUpdateDoublesSketch.java
@@ -217,7 +217,7 @@ final class DirectUpdateDoublesSketch extends DirectUpdateDoublesSketchR {
   //Restricted overrides
 
   @Override
-  HeapUpdateDoublesSketch getSketchAndReset() {
+  UpdateDoublesSketch getSketchAndReset() {
     final HeapUpdateDoublesSketch skCopy = heapify(this);
     reset();
     return skCopy;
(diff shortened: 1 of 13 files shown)


src/main/java/org/apache/datasketches/quantiles/DirectUpdateDoublesSketch.java
```diff
@@ -99,6 +100,11 @@ static DirectUpdateDoublesSketch newInstance(final int k, final MemorySegment ds
     return new DirectUpdateDoublesSketch(k, dstSeg, mSegReq);
   }
 
+  static HeapUpdateDoublesSketch heapify(final DirectUpdateDoublesSketch skIn) {
+    final MemorySegment segIn = skIn.getMemorySegment();
```
freakyzoidberg, line 104: nit, we could avoid some debugging headaches here.

```
Objects.requireNonNull(skIN, "skIn must be non-null");
```

leerho: This method is package-private. Looking at the call hierarchy, this method is _only_ called from a method in the same class:

```
  UpdateDoublesSketch getSketchAndReset() {
    final HeapUpdateDoublesSketch skCopy = heapify(this);
    ...
```
Where the parameter passed is the owning class, which has to exist. So it can never be null. 

Nonetheless, your observation is fair, and applies generally to the whole library.  My preferred practice is to check parameter validity on the public methods where the parameters originate, or at least higher up the call stack, if possible, rather than on the lower level internal methods.  But this code has multiple authors, which make it difficult to enforce.

I think it is OK to leave this particular method the way it is. But come back later with a strategy, either through testing or more careful design across the entire library to avoid difficult to debug null pointers. 

Thank you for your observation!

PR #687, opened 2025-10-03
Keep comment cases consistent

tisonkun:


src/main/java/org/apache/datasketches/hll/TgtHllType.java
```diff
@@ -60,7 +60,7 @@ public enum TgtHllType {
    */
   HLL_6,
   /**
-   * An Hll Sketch with a bin size of 8 bits
```
leerho, line 63: Really?  The HyperLogLog abbreviation appears in multiple cases both in code and comments:  hll, Hll, HLL.  And, yes, this could have been spelled all caps as HLL, but I don't think that it should cause any confusion.  

I appreciate your wanting to make a contribution, thank you, but it would be even more appreciated if you could find more substantive areas where improved comments would add clarity, and I am sure there are many :)

tisonkun: OK.

This is the only one different in the three variants.

I go through the code and touch this on the way.

If this is no the contribution you expect, I can close this one.

tisonkun: By the way, I filed an issue at https://github.com/apache/datasketches-java/issues/688 when reading the code, I'll appreciate it if you can give a look :D

tisonkun: As the background, here is how I easily find this inconsistency:

IMG_6538

But I understand different projects have different preference. Thanks for your reply >~<

leerho: OK, I corrected this in my current PR that is waiting for review. 
Thanks.

PR #704, opened 2025-12-19
Accept TDigest bytes with buffered values

### LINKED ISSUE #701: Serde incompatible with C++ TDigest
In Java's TDigestDouble, this field is always ignored as unused.

https://github.com/apache/datasketches-java/blob/37edd8c9898f8f073aaafad7097cb0949eb84c85/src/main/java/org/apache/datasketches/tdigest/TDigestDouble.java#L429

However, C++'s impl can use it for denote the size of buffer vector.

https://github.com/apache/datasketches-cpp/blob/7bb979d3ef8929e235bcd22d67579e1f695f6ecd/tdigest/include/tdigest_impl.hpp#L321-L326

Also the Java impl always assumes there is no buffer serialized. This can lead to serde incompatible for any C++ TDigest has buffered data serialized with `with_buffer`, like:

https://github.com/apache/datasketches-cpp/blob/7bb979d3ef8929e235bcd22d67579e1f695f6ecd/tdigest/test/tdigest_test.cpp#L222-L235

tisonkun:
This closes #701.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,28 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+root = true
+
+[*]
+end_of_line = lf
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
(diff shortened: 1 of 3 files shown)


.editorconfig
tisonkun, line 1: Help IDEA to properly format files without large indentation changes. Can be removed if you don't like to check it into the VCS but IMO it does no harm.

tisonkun: Out of the thread: spotless can help in autoformat over checkstyle

See also:

* https://github.com/apache/curator/pull/463
* https://github.com/diffplug/spotless/tree/main/plugin-maven

src/test/java/org/apache/datasketches/tdigest/TDigestCrossLanguageTest.java
```diff
@@ -25,54 +25,68 @@
...
-        assertEquals(td.getRank(n + 1), 1);
-        if (n == 1) {
-          assertEquals(td.getRank(n), 0.5);
+    final boolean[] with_buffer = {false, true};
+    for (final boolean buffered : with_buffer) {
+      final int[] nArr = {0, 1, 10, 100, 1000, 10_000, 100_000, 1_000_000};
+      for (final int n : nArr) {
+        final byte[] bytes;
+        if (buffered) {
+          bytes = Files.readAllBytes(cppPath.resolve("tdigest_double_buf_n" + n + "_cpp.sk"));
```
tisonkun, line 43: This will fail without this patch.

PR #708, opened 2026-01-01
Fix BinomialBoundsN may divide 0

### LINKED ISSUE #707: BinomialBoundsN may divide 0
From exception, BinomialBoundsN is expect to not take the 0 theta.(To prevent "divide 0" happend). https://github.com/apache/datasketches-java/blob/b1d07af37eaa508915bd40559184974bf424f837/src/main/java/org/apache/datasketches/thetacommon/BinomialBoundsN.java#L276  But actually it doesn't check whether theta is 0. 

Following test is fail to check this because this test will pass if any `checkArgs` throw exception. 
```
  @Test
  public static void checkCheckArgs() {
    try {
      checkArgs(-1L, 1.0, 1);
      checkArgs(10L, 0.0, 1);
      checkArgs(10L, 1.01, 1);
      checkArgs(10L, 1.0, 3);
      checkArgs(10L, 1.0, 0);
      checkArgs(10L, 1.0, 4);
      fail("Expected SketchesArgumentException");
    } catch (final SketchesArgumentException e) {
      //pass
    }
  }
```

ZENOTME:
Fix #707

### DIFF
--- a/src/test/java/org/apache/datasketches/thetacommon/BinomialBoundsNTest.java
+++ b/src/test/java/org/apache/datasketches/thetacommon/BinomialBoundsNTest.java
@@ -25,6 +25,7 @@ import static org.apache.datasketches.thetacommon.BinomialBoundsN.getUpperBound;
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertTrue;
 import static org.testng.Assert.fail;
+import static org.testng.Assert.assertThrows;
 
 import org.apache.datasketches.common.SketchesArgumentException;
 import org.apache.datasketches.thetacommon.BinomialBoundsN;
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/datasketches/thetacommon/BinomialBoundsNTest.java
```diff
@@ -22,9 +22,7 @@
 import static org.apache.datasketches.thetacommon.BinomialBoundsN.checkArgs;
 import static org.apache.datasketches.thetacommon.BinomialBoundsN.getLowerBound;
 import static org.apache.datasketches.thetacommon.BinomialBoundsN.getUpperBound;
-import static org.testng.Assert.assertEquals;
-import static org.testng.Assert.assertTrue;
-import static org.testng.Assert.fail;
+import static org.testng.Assert.*;
```
leerho: Starred imports are discouraged as they create namespace pollution.  Especially here, as the number of potential imports under _org.testng.Assert.*_ is quite large.
Please return to the specific imports.
Thanks.

leerho (changes requested):
Thanks for catching the incomplete check on theta!

src/test/java/org/apache/datasketches/thetacommon/BinomialBoundsNTest.java
ZENOTME: Fixed

tisonkun (approved):
Thanks for your contribution! LGTM.

PR #711, opened 2026-01-07
Allow override surefire plugin group

tisonkun:
Previously, the exclude all groups config is defined in `/plugins/plugin` and thus it overrides any other configs defined in the `/pluginManagement` section.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -129,6 +129,20 @@ under the License.
   </dependencies>
 
   <build>
+      <pluginManagement>
+          <plugins>
+              <plugin>
+                  <!-- Apache Parent pom, pluginManagement-->
+                  <groupId>org.apache.maven.plugins</groupId>
+                  <artifactId>maven-surefire-plugin</artifactId>
+                  <configuration>
+                      <!-- by default, exclude all snapshot targets -->
+                      <!-- this will be overridden on dedicated profile -->
+                      <excludedGroups>${testng.generate-java-files},${testng.check-cpp-files},${testng.check-go-files},${testng.check-cpp-historical-files}</excludedGroups>
+                  </configuration>
+              </plugin>
+          </plugins>
+      </pluginManagement>
       <plugins>
 
       <plugin>
@@ -143,7 +157,7 @@ under the License.
           <fork>true</fork>
           <release>${java.version}</release>
           <compilerArgs>
-            <arg>-J${jvm.options}</arg> <!-- comma separated or separate args --> 
+            <arg>-J${jvm.options}</arg> <!-- comma separated or separate args -->
           </compilerArgs>
         </configuration>
       </plugin>
@@ -224,7 +238,7 @@ under the License.
           <doclint>all,-missing</doclint>
           <release>${java.version}</release>
           <additionalJOptions> <!-- requires -J prefix -->
-            <additionalJoption>-J${jvm.options}</additionalJoption> 
+            <additionalJoption>-J${jvm.options}</additionalJoption>
         </additionalJOptions>
         </configuration>
         <executions>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -274,8 +288,6 @@ under the License.
           <useManifestOnlyJar>false</useManifestOnlyJar>
           <redirectTestOutputToFile>true</redirectTestOutputToFile>
           <reportsDirectory>${project.build.directory}/test-output/${maven.build.timestamp}</reportsDirectory>
-          <excludedGroups>${testng.generate-java-files},${testng.check-cpp-files},${testng.check-go-files},
-${testng.check-cpp-historical-files}</excludedGroups> <!-- do not indent -->
```
leerho, line 278: I had set this up so that an ordinary "mvn test" would not generate the java files and run the tests against the other languages.  These are run with specific profiles.   

Now that we are adding more languages into the mix, I am not clear on what you have in mind.

pom.xml
```diff
@@ -129,6 +129,20 @@ under the License.
...
+                  <groupId>org.apache.maven.plugins</groupId>
+                  <artifactId>maven-surefire-plugin</artifactId>
+                  <configuration>
+                      <!-- by default, exclude all snapshot targets -->
+                      <!-- this will be overridden on dedicated profile -->
+                      <excludedGroups>${testng.generate-java-files},${testng.check-cpp-files},${testng.check-go-files},${testng.check-cpp-historical-files}</excludedGroups>
+                  </configuration>
+              </plugin>
+          </plugins>
+      </pluginManagement>
```
leerho, line 145: The ds-java POM does not have any maven modules so there is no inheritance.  I also was warned that a configuration specification in a plugin under pluginManagement may be ignored by some plugins.  Also this configuration will be overridden by the surefire plugin specified under \<build\>\<plugins\>.  So what  is achieved by putting this plugin under pluginManagement with a different configuration?

leerho (changes requested):
I have a couple of questions that need clarification.

pom.xml
tisonkun: Note that we use `/pluginManagement/plugins/plugin` to override the exclude group of surefire plugin:

```xml
    <profile>
      <id>generate-java-files</id>
      <build>
        <pluginManagement>
          <plugins>
            <plugin>
              <!-- Apache Parent pom, pluginManagement-->
              <groupId>org.apache.maven.plugins</groupId>
              <artifactId>maven-surefire-plugin</artifactId>
              <configuration>
                <groups>${testng.generate-java-files}</groups>
                <excludedGroups>${testng.check-cpp-files},${testng.check-go-files},${testng.check-cpp-historical-files}</excludedGroups>
              </configuration>
            </plugin>
          </plugins>
        </pluginManagement>
      </build>
    </profile>
```

So this PR changes all the configurations to `/pluginManagement` for proper overriding.

> configuration will be overridden by the surefire plugin specified under <build><plugins>.

Yes. The current `pom.xml` before this patch, would always use:

```xml
 <excludedGroups>${testng.generate-java-files},${testng.check-cpp-files},${testng.check-go-files},
${testng.check-cpp-historical-files}</excludedGroups> <!-- do not indent -->
```

Because it is the final config. Thus, any profile won't be applied actually.

tisonkun: An alternative is to always use `/plugins/plugin` for configuring the surefire plugin in every profile.

pom.xml
tisonkun: See https://github.com/apache/datasketches-java/pull/711#discussion_r2670368334.

This config would be still applied, but from the `/pluginManagement/plugins/plugin` section that can be override.

This PR is a no change for developer who only runs `mvn test`.

leerho (approved):

PR #712, opened 2026-01-07
CI workflows should run if pom.xml or workflow yaml files changed

tisonkun:
Any dedicated unrelated files should be ignored more concretely.

Currently, you can see https://github.com/apache/datasketches-java/pull/711 doesn't trigger CI workflows while it should do.

### DIFF
--- a/.github/workflows/javadoc.yml
+++ b/.github/workflows/javadoc.yml
@@ -41,38 +41,38 @@ jobs:
           TAG_NAME: ${{ github.event.inputs.tag_ref }}
         run: |
           if [ -z "$TAG_NAME" ]; then echo "ERROR: No tag specified"; exit 1; fi
-          
+
           # 1. Initialize error tracking
           EXIT_CODE=0
-          
+
           # 2. Configure Git Identity
           git config user.email "noreply@github.com"
           git config user.name "github-actions[bot]"
-          
+
           # 3. Ensure gh-pages exists and is fetched
           echo "ECHO: git fetch origin gh-pages"
           git fetch origin gh-pages
-          
+
           # 4. Create worktree for the gh-pages branch in a separate folder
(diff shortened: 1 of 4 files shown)


.github/workflows/javadoc.yml
```diff
@@ -41,38 +41,38 @@ jobs:
           TAG_NAME: ${{ github.event.inputs.tag_ref }}
         run: |
           if [ -z "$TAG_NAME" ]; then echo "ERROR: No tag specified"; exit 1; fi
-          
+
```
leerho, line 44: This is strange.  I've been having a lot of GHA parser errors and I read that all lines under a "run: |" must have the same indentation, including blank lines, or it will confuse the parser.  Are you sure this will work?
Did you run it?

.github/workflows/javadoc.yml
leerho, line 98: Just a comment.  Originally I had the echo command right after the "run:", but the "ECHO: " inside the quotes confused the parser thinking it was a GHA command.  It took me hours to find this and multiple rounds with Gemini was no help.  Moving the echo under a "run: |" solved the problem.  The GHA parser is extremely fussy.

I start the echo commands with "ECHO:" so I can easily find them in the trove of output.

.github/workflows/javadoc.yml
tisonkun: I guess, if you have some indentation, they should keep the same; if you have a blank line without any whitespace, the parser can properly recognize it as a blank line.

Semantic indentation is terrible :/

.github/workflows/javadoc.yml
tisonkun: I suppose this is a YAML issue:

https://yaml.org/spec/1.2.2/#literal-style

tisonkun: YAML and Windows (`shell: bash` etc.) are most of the error sources from CI workflows :/

PR #722, opened 2026-01-30
chore: tidy comment and code of CpcSketch

tisonkun:


src/main/java/org/apache/datasketches/cpc/CpcUnion.java
```diff
@@ -124,7 +124,7 @@ public CpcUnion(final int lgK, final long seed) {
     bitMatrix = null;
     // We begin with the accumulator holding an EMPTY_MERGED sketch object.
     // As an optimization the accumulator could start as NULL, but that would require changes elsewhere.
-    accumulator = new CpcSketch(lgK);
+    accumulator = new CpcSketch(lgK, seed);
```
tisonkun, line 127: Can be a bug fix.

PR #727, opened 2026-02-18
Cleanup read and write bytes in test & pom improvements

leerho:
This replaces a bunch of methods attempting to read file bytes. These methods were not robust, especially in a Windows OS environment, and were also creating memory leaks.  This solution is far simpler and more robust and modern solution that is only one method.

This also fixes the methods used to write bytes to a file that had similar problems with Windows.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,4 @@ _*/
 
 # Sketch binary test files
 *.sk
+/tools/genCPPcrossLangFiles.sh
--- a/pom.xml
+++ b/pom.xml
@@ -99,10 +99,9 @@ under the License.
     <project.reporting.outputDirectory>${project.build.directory}/site</project.reporting.outputDirectory>
     <maven.build.timestamp.format>yyyy-MM-dd'T'HH-mm-ss'Z'</maven.build.timestamp.format>
 
-    <!--  Covered by Apache Parent POM -->
     <!--    org.apache.maven plugins -->
     <maven-assembly-plugin.version>3.8.0</maven-assembly-plugin.version>
-    <maven-compiler-plugin.version>3.15.0</maven-compiler-plugin.version> <!-- parent 37: 3.14.1 -->
+    <maven-compiler-plugin.version>3.15.0</maven-compiler-plugin.version>
     <maven-deploy-plugin.version>3.1.4</maven-deploy-plugin.version>
     <maven-enforcer-plugin.version>3.6.2</maven-enforcer-plugin.version>
     <maven-gpg-plugin.version>3.2.8</maven-gpg-plugin.version>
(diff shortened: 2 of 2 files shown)


src/test/java/org/apache/datasketches/common/TestUtilTest.java
```diff
@@ -0,0 +1,66 @@
...
+  public void testGetFileBytes_MissingFile() {
+    byte[] resultBytes = getFileBytes(resPath, "NonExistentFile");
+    assertNotNull(resultBytes);
+    assertEquals(resultBytes.length, 0, "Should return empty array for missing file.");
+  }
+
+  @Test
+  public void testGetFileBytes_NotRegular_NotReadable() throws IOException {
+    try {
+      byte[] resultBytes = getFileBytes(resPath, "");
```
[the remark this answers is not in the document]
leerho: Fixed.

src/test/java/org/apache/datasketches/common/TestUtil.java
```diff
@@ -40,227 +42,98 @@
...
-        throw new IllegalStateException("Failed to set only owner 'Writable' on file");
-      }
-
-    } catch (final IOException e) { throw new RuntimeException(e); }
-    return file;
+      Files.createDirectories(basePath); //create the directory if it doesn't exist.
+      filePath = basePath.resolve(fileName);
+      Files.write(filePath, bytes);
+    } catch (IOException e) {
+      throw new RuntimeException("System IO Error writing file: " + filePath.toString() + " " + e);
```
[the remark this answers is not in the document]
leerho: Fixed

src/test/java/org/apache/datasketches/common/TestUtil.java
```diff
@@ -19,248 +19,122 @@
...
-   * @throws IllegalArgumentException if array size grows larger than what can be safely allocated by some VMs.
-
+   * Gets all the bytes of a file as a byte array.
+   * If the file is missing, this either throws an exception or writes a warning message to the console
+   * based on the state of the optional {@link #Existence Existence}.
+   * @param basePath the base directory path where the file is located
+   * @param fileName the simple file name of the file
+   * @param option an optional parameter. If option == Existence.MUST_EXIST and the file does not exist an exception will be thrown.
+   * If option == Existence.WARNING, or not given, and the file does not exist, it writes a warning message
+   * to {@link System.err.out System.err.out}.
```
proost: typo? System.err System.out

src/test/java/org/apache/datasketches/common/TestUtil.java
```diff
@@ -19,248 +19,122 @@
...
-import java.util.Arrays;
-import java.util.List;
 import java.util.Objects;
 
 /**
  * Utilities common to testing
  */
 public final class TestUtil  {
 
-  private static final String userDir = System.getProperty("user.dir");
```
proost, line 43: When run some test cases in the IDE(In my case, IntelliJ), because of this change, failed. 

This change is good but but I think just mentioning about this in thre README.md or somewhere is enough for local development.

leerho: I don't understand.  Are you saying that InteliJ doesn't understand the construct: Path.of(".", "...", "...")?

src/test/java/org/apache/datasketches/common/TestUtil.java
leerho: Oops, should be just System.err.  Fixed.

src/test/java/org/apache/datasketches/common/TestUtil.java
leerho: OK, I did some research, and changed the path resolution to a more robust method (I hope).

Please try again  :)

proost: Thank you! it works well

PR #733, opened 2026-04-10
fix: inconsistency in sorted view

proost:
I try to make consistency in docs and implementation.

1. Docs said raising IllegarArgumentException when empty, but actual is SketchesArgumentException. 
2. getMaxItem and getMinItem not raise expected exception.


leerho (approved):
Thank you!

PR #734, opened 2026-04-25
fix: respect negative values on deserialization in REQ sketch

proost:
`Float.MIN_VALUE` is smallest positive value. So if REQ sketch include negative values, min value is not respected after deserialization.

C++ is ok, because it delegates to comparator not like java.

After this PR merged, I will send PR for compat in C++.

### DIFF
--- a/src/main/java/org/apache/datasketches/req/ReqSerDe.java
+++ b/src/main/java/org/apache/datasketches/req/ReqSerDe.java
@@ -205,7 +205,7 @@ class ReqSerDe {
     final float[] arr = new float[count];
     posSeg.getFloatArray(arr, 0, count);
     float minItem = Float.MAX_VALUE;
-    float maxItem = Float.MIN_VALUE;
+    float maxItem = -Float.MAX_VALUE;
     for (int i = 0; i < count; i++) {
       minItem = min(minItem, arr[i]);
       maxItem = max(maxItem, arr[i]);
--- a/src/main/java/org/apache/datasketches/req/ReqSerDe.java
+++ b/src/main/java/org/apache/datasketches/req/ReqSerDe.java
@@ -19,14 +19,16 @@
 
 package org.apache.datasketches.req;
 
+import static java.lang.Math.max;
+import static java.lang.Math.min;
+import static java.lang.Math.round;
+
 import java.lang.foreign.MemorySegment;
 import java.util.ArrayList;
 import java.util.List;
 
 import org.apache.datasketches.common.Family;
 import org.apache.datasketches.common.positional.PositionalSegment;
-
-import static java.lang.Math.*;
 
 /**
  * This class handles serialization and deserialization.
@@ -205,11 +207,6 @@ class ReqSerDe {
     float minItem = Float.POSITIVE_INFINITY;
     float maxItem = Float.NEGATIVE_INFINITY;
     for (int i = 0; i < count; i++) {
-      final float item = arr[i];
-      if (Float.isNaN(item)) {
-        continue;
-      }
-
       minItem = min(minItem, arr[i]);
       maxItem = max(maxItem, arr[i]);
     }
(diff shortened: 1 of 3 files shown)


leerho (changes requested):
See comment in the code.

src/main/java/org/apache/datasketches/req/ReqSerDe.java
```diff
@@ -205,7 +205,7 @@ static final Compactor extractCompactor(final PositionalSegment posSeg, final bo
     final float[] arr = new float[count];
     posSeg.getFloatArray(arr, 0, count);
     float minItem = Float.MAX_VALUE;
-    float maxItem = Float.MIN_VALUE;
+    float maxItem = -Float.MAX_VALUE;
```
leerho: What you found here is obviously incorrect, but neither is your solution.  Internally, for the Classic Quantiles, KLL and REQ, the min and max values are initialized to NaN. and the first incoming value that is not NaN, sets both the min and max values. This allows input values of +/- infinity, which your solution would not recognize.  I would argue that you should use a similar solution here, for consistency if nothing else. 

I would also suggest you create a test that uses +/- Infinities, just to confirm that it works.

proost: [d67731a](https://github.com/apache/datasketches-java/pull/734/commits/d67731a309a28e2c7bad4571b9c318c5599476b0)

Oh, I was wrong. I changed like other sketches or Update method. 

And also add Infinity cases

leerho (changes requested):
See new comment inline.

src/main/java/org/apache/datasketches/req/ReqSerDe.java
```diff
@@ -204,11 +203,17 @@ static final Compactor extractCompactor(final PositionalSegment posSeg, final bo
...
-      minItem = min(minItem, arr[i]);
-      maxItem = max(maxItem, arr[i]);
+      final float item = arr[i];
+      if (Float.isNaN(minItem)) {
+        minItem = item;
+        maxItem = item;
+      } else {
+        if (item < minItem) { minItem = item; }
+        if (item > maxItem) { maxItem = item; }
+      }
```
leerho: I have taken a closer look at this, and what you have here is not correct either.  If a NaN should happen to appear, it would reset both the minItem and maxItem values to NaN, erasing the tracked minItem and maxItem values.

This extractCompactor is part of the deserialization process, and since NaNs are rejected during sketch construction, they should not appear in the serialization data.  Nonetheless, this section of code is just trying to reconstruct the min and max values of this specific compactor. 

These specific min and max values here are only used in a sketch that is in EXACT format, which has a single compactor.  The min and max values are not stored in the serialization structure itself and must be derived dynamically from the compactor.  This keeps the overall storage small for small data streams.

Note that in the normal, ESTIMATION format the min and max values are stored in the serialization structure itself, so the min and max values being computed here are not used. (This means the error you observed where it looses negative values probably occurred with a very small data set.)

I think we can trust that the values being examined from the serialized data structure do not contain NaNs.   This means that to correct the original error, all we need to do is correct the initialization of the min and max values to +/- infinity.  So the corrected section should look like the following, which will preserve any infinities if they exist:

```
    posSeg.getFloatArray(arr, 0, count);
    float minItem = Float.POSITIVE_INFINITY;   // changed line
    float maxItem = Float.NEGATIVE_INFINITY; // changed line 
    for (int i = 0; i < count; i++) {
      minItem = min(minItem, arr[i]);
      maxItem = max(maxItem, arr[i]);
    }
```

leerho (changes requested):

src/main/java/org/apache/datasketches/req/ReqSerDe.java
proost: fixed it to follow the code.

leerho (approved):

PR #737, opened 2026-06-30
Add HllUnion.update(HllUnion)

leerho:
This adds the ability of the HllUnion operator to update from HllUnions directly, bypassing an expensive copy.

Corrects some spelling errors.

### DIFF
--- a/src/test/java/org/apache/datasketches/hll/UnionCaseTest.java
+++ b/src/test/java/org/apache/datasketches/hll/UnionCaseTest.java
@@ -31,8 +31,8 @@ import static org.testng.Assert.assertFalse;
 import static org.testng.Assert.assertTrue;
 
 import java.lang.foreign.MemorySegment;
-import java.lang.invoke.MethodHandles;
-import java.lang.invoke.VarHandle;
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/datasketches/hll/UnionCaseTest.java
```diff
@@ -52,48 +54,68 @@ public class UnionCaseTest {
...
     final HllUnion union = getUnion(caseNum, gdtSeg);
-    union.update(source);
+    if (srcUnion) {
+      union.update(uSource);
+    } else {
+      union.update(skSource);
+    }
     final int totalU = getSrcCount(caseNum, maxLgK) + getUnionCount(caseNum);
-    output(caseNum, source, union, totalU);
+    output(caseNum, skSource, union, totalU);
```
proost: Looks like  `skSource` -> `srcUnion ? uSource : skSource` correct.

tisonkun (approved):
LGTM.

cc @notfilippo This may benefit rust impl's HllUnion update as well. But Rust doesn't have override so we may consider the method name.

Issue #702, opened 2025-12-17
Compress TDigest can generate centroid with NaN mean


tisonkun:
See this thread https://github.com/apache/datasketches-rust/pull/23#discussion_r2622754228

Also we don't verify that the centroid mean is not NaN and the weight is not zero (should be positive even?) on deserialization. (Generally we can't assume the data to be deserialized is safe.)

I think the rank/quantile method can return NaN if the calculation result in that. But the stored centroid should not be NaN or else stable sort is unsound?

leerho:
Sounds like this should be a bug fix for both C++ and Java.

leerho:
@tisonkun 
Where are we with this issue?  There have been a number of PRs affecting TDigest, have any of them addressed this correctly so we can close this issue?

tisonkun:
I'll check this issue this week.

PR #742, opened 2026-07-15
Prevent NaN centroid means in TDigestDouble (#702)

PerumalsamyR:
Addresses #702

### Problem

When merging a value into an existing centroid, the mean was updated incrementally:

```java
centroidMeans_[i] += ((values[current] - centroidMeans_[i]) * weights[current]) / centroidWeights_[i];
```

The intermediate `(value - mean)` (or its product with the weight) can overflow to infinity even when both operands are finite — e.g. means near opposite ends of the double range. Once a centroid mean becomes infinite, a subsequent update produces `Inf + (-Inf) = NaN`, which is then stored, breaking the sort/binary-search invariants and getting serialized. The same overflow pattern existed in `weightedAverage()`, used for quantile interpolation.

This mirrors the fix already applied to the Rust implementation in apache/datasketches-rust#23.

### Changes

- `merge()`: keep the incremental mean update in the common path (bit-identical results), but fall back to an overflow-safe weighted average `m1*(w1/w) + m2*(w2/w)` when the result is not finite. Each term of that form is bounded by the magnitude of its input, so the stored mean stays finite for finite inputs.
- `weightedAverage()`: normalize the weights before multiplying so quantile interpolation between centroids of large magnitude cannot overflow.
- `update()`: ignore ±Infinity in addition to NaN. With mixed infinities in the input, no update formula can keep centroid means well-defined, and rejecting them was the approach endorsed in the Rust PR discussion.
- `heapify()` / compat formats: validate deserialized state as requested in the issue — min/max, centroid means and buffered values must be finite, centroid weights must be positive — throwing `SketchesArgumentException` otherwise, since serialized input cannot be assumed to be well-formed.

### Testing

- New test `extremeValuesDoNotProduceNaN` feeds 10k alternating `±Double.MAX_VALUE` values: on current main it produces non-finite quantiles (and NaN centroid means); with this change all centroid means and quantiles stay finite. 
- New tests for `update()` ignoring NaN/±Inf and for `heapify()` rejecting corrupted bytes (NaN/Inf mean, NaN min, zero weight, NaN single value).
- Full suite passes: `mvn clean test` — 2174 tests, 0 failures.

### DIFF
--- a/src/main/java/org/apache/datasketches/tdigest/TDigestDouble.java
+++ b/src/main/java/org/apache/datasketches/tdigest/TDigestDouble.java
@@ -95,11 +95,12 @@ public final class TDigestDouble {
   }
 
   /**
-   * Update this TDigest with the given value
+   * Update this TDigest with the given value.
+   * NaN and infinity are ignored.
    * @param value to update the TDigest with
    */
   public void update(final double value) {
-    if (Double.isNaN(value)) { return; }
+    if (!Double.isFinite(value)) { return; }
     if (numBuffered_ == (centroidsCapacity_ * BUFFER_MULTIPLIER)) { compress(); }
     bufferValues_[numBuffered_] = value;
     numBuffered_++;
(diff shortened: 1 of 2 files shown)


leerho (approved):
LGTM

PR #757, opened 2026-09-02
Fix CpcWrapper throwing on empty sketches

davecromberge:
`CpcWrapper.getEstimate()`, `getLowerBound()` and `getUpperBound()` throw
`SketchesStateException` for any empty sketch, so the class can't be used by a caller
whose data may contain one.

An empty image carries none of the high preamble fields. `EMPTY_HIP` sets the HIP flag, so
`getEstimate()` takes the HIP branch and asks for `HIP_ACCUM`; `EMPTY_MERGED` takes the ICON
branch and asks for `NUM_COUPONS`. Neither field is present, so both throw.

Empty now reports zero coupons and a zero HIP accumulator, matching `CpcSketch`.

Added `CpcWrapperEquivalenceTest`, which checks the wrapper against a heapified sketch on
estimate, both bounds and lgK, across every lgK, the coupon counts that drive the flavor
transitions, union results, and the three ways a sketch can be empty.

### DIFF
--- a/src/main/java/org/apache/datasketches/cpc/CpcWrapper.java
+++ b/src/main/java/org/apache/datasketches/cpc/CpcWrapper.java
@@ -65,15 +65,29 @@ public final class CpcWrapper {
     this(MemorySegment.ofArray(byteArray));
   }
 
+  //An empty image carries none of the high preamble fields, so they must not be read.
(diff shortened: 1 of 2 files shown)


leerho (approved):
LGTM

PR #765, opened 2026-09-15
Fix TDigest extreme means and validate serialized state

### LINKED ISSUE #702: Compress TDigest can generate centroid with NaN mean

tisonkun:
Fixes #702.

`TDigestDouble` can still store an infinite centroid mean when merging finite, same-sign extreme values with large weights. The incremental calculation overflows and its fallback can also overflow because it independently rounds both normalized weights. Quantile interpolation has the same issue, and a deserialized last centroid with weight 2 can produce `NaN` through division by zero.

Normalize the merge weight before multiplication and use fused interpolation when the difference is finite. For quantile interpolation, use a sign-aware weighted average; handle the two-sample right-tail boundary explicitly.

Also complete deserialization validation across native double/float and reference double/float formats:

- Reject fractional, non-finite, or out-of-range reference weights before narrowing to `long`.
- Reject overflowing total weights, including buffered values. Reject overflowing updates and merges before changing the destination.
- Check flags, counts, payload lengths, ordered extrema, centroid ordering, and centroid/buffer membership in the stored range before accepting the image.
- Accept valid images with more centroids or buffered values than local compression thresholds. Preserve compact centroids on round-trip and compress an overfull buffer before the next update.
- Preserve valid reference compact images: their float means can round outside double extrema, so validate in float precision and clamp those means to the original bounds.

## Rust comparison

Reviewed the T-Digest source history through Rust main `e85b5888989f9329e375bc3410aa619d9881a53d`, including correctness fixes embedded in buffer optimizations and API work.

| Rust fix | Java status |
| --- | --- |
| Non-finite inputs and stored values ([apache/datasketches-rust#23](https://github.com/apache/datasketches-rust/pull/23)) | Already covered on main; retain the validation. |
| Repeated-value singleton interpolation boundary ([apache/datasketches-rust#63](https://github.com/apache/datasketches-rust/pull/63)) | Main already uses `<= 0.5`; add the exact repeated-value regression. |
| Stable centroid arithmetic and weight overflow ([apache/datasketches-rust#88](https://github.com/apache/datasketches-rust/pull/88), [apache/datasketches-rust#173](https://github.com/apache/datasketches-rust/pull/173)) | Completed here, including overflow checks before updates and merges mutate state. |
| Finite quantiles for both signs of extreme values ([apache/datasketches-rust#173](https://github.com/apache/datasketches-rust/pull/173)) | Completed here with sign-aware interpolation. |
| Overfull deserialized buffers and retained compact centroids ([apache/datasketches-rust#210](https://github.com/apache/datasketches-rust/pull/210), [apache/datasketches-rust#231](https://github.com/apache/datasketches-rust/pull/231)) | Completed here; local sizing thresholds do not constrain the wire format. |
| Empty CDF/PMF split lists ([apache/datasketches-rust#247](https://github.com/apache/datasketches-rust/pull/247)) | Already valid in Java; add the one-bin regression. |
| Swapped interior interpolation weights ([apache/datasketches-rust#250](https://github.com/apache/datasketches-rust/pull/250)) | Already fixed by `a2ab8cf41`; add exact directional interpolation assertions. |
| Right quantile tail sign and left rank tail normalization ([apache/datasketches-rust#250](https://github.com/apache/datasketches-rust/pull/250)) | Already fixed by `a2ab8cf41` and `642fbdfa6`; add exact symmetric rank, quantile, CDF, and PMF assertions. |
| Weight-two right-tail division by zero ([apache/datasketches-rust#250](https://github.com/apache/datasketches-rust/pull/250)) | Completed here. |
| Serialized state invariants ([apache/datasketches-rust#262](https://github.com/apache/datasketches-rust/pull/262)) | Completed here across all four input formats. |

Java already promotes `k * 2` to `int`, so Rust's widened multiplication from [apache/datasketches-rust#209](https://github.com/apache/datasketches-rust/pull/209) needs no Java change. Rust-specific ownership, API, and allocation refactors are not required for these correctness fixes.

## Validation

- Confirmed regression failures against the unchanged main implementation, including an infinite stored centroid mean, infinite quantile, and `NaN` at the two-sample right tail.
- Verified that restoring each old weight-order, tail-sign, tail-normalization, singleton-boundary, or buffer-threshold bug makes its focused regression fail.
- `mvn -B clean verify`: 2,265 tests passed, including the new numeric and malformed-input regressions; Javadoc and RAT checks passed.
- `TDigestCrossLanguageTest`: all 5 selected test methods passed against 32 C++, 16 Go, 8 Java, and 2 reference TDigest snapshots. The cross-language snapshots came from DataSketches TCK revision `c0a180708c6e6433e4cba7fba091713eb8af3eaa`.

```sh
mvn -B test -Pcheck_all_language_files \
  -Dtest=TDigestCrossLanguageTest \
  -Dtestng.all_groups=check_cpp_files,check_go_files,check_java_files,check_cpp_historical_files \
  -Dtest.data.root=/tmp/issue702-compat
```


leerho (approved):
Thank you!

PR #769, opened 2026-09-16
fix(tdigest): preserve stored min and max when merging

jaideeppyne:
## Summary

- Copy the stored min/max from the other digest in `TDigestDouble.merge` before compressing the combined centroid list.
- Add a regression that heapifies an image whose stored min/max sit strictly outside the extreme centroid means, then merges it into an empty digest and into a non-empty one.

## Why

`merge` previously re-derived extrema from the extreme centroid means. That is exact only when those centroids have unit weight. Deserialized images (and the reference implementation format) can have weighted tails, so min/max drifted inward and rank/quantile tail interpolation used the wrong endpoints.

Same defect as apache/datasketches-rust#277. Independent of #767.

## Validation

- `TDigestDoubleTest#mergePreservesDeserializedMinMaxWithWeightedTails`


leerho (approved):
Thank you for uncovering and fixing this!

